PR #2, opened 2018-11-23
Allow registering JSS for server side rendering

nickbrowne:
Adds some optional props to `ThemeProvider` to support server side rendering.

Based on:
https://material-ui.com/guides/server-rendering/
https://github.com/cssinjs/jss/blob/master/docs/ssr.md

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
     "prop-types": "^15.6.2",
     "react": "^16.6.3",
     "react-dom": "^16.6.3",
+    "react-jss": "^8.6.1",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -42,6 +42,7 @@
     "prop-types": "^15.6.2",
     "react": "^16.6.3",
     "react-dom": "^16.6.3",
+    "react-jss": "^8.6.1",
```
nullobject: This should probably be a regular dependency, right?

PR #4, opened 2018-12-04
TC-393 Add button component documentation

nullobject:
This is an alternative to the info addon approach in #3, here we explicitly generate all the documentation ourselves using the [storybook readme addon](https://github.com/tuchk4/storybook-readme). In this PR, I have only documented the `<Button>` component.

IMHO this is much simpler than using the info addon, because you are just writing raw markdown. I have tried to start a convention of keeping the stories in separate files,
 and embedding the markdown along with the story, so that it's all in one place. Hopefully this strikes a good balance between keeping things separate vs. reducing friction for devs to write new components.

I also created a `<ComponentOverview>` component, which will render an overview of a react component based on the [docgen](https://github.com/storybooks/babel-plugin-react-docgen) data. This includes the class description and the prop definitions.

We end up with something that looks like this:

screen shot 2018-12-05 at 4 17 05 pm
screen shot 2018-12-05 at 4 16 54 pm

Things I like better than the `info` addon:

- You can have whatever you want in the docs (it's just markdown).
- You don't have to do avoid certain components in your story to get the prop types and example source code to look correct (e.g. `<WithStyles(Grid) item>`).

Things that the `info` addon does better:

- The prop definitions include the values of all enums, etc. This is something we can probably implement in a followup PR, I just thought I'd keep things fairly simple for now.


nullobject:
> Hmm, it does support a level of custom text:

Great, it might be worth a play. I do like that the info addon is one of the official addons. The readme one is a community addon, so it may see less active development.

src/Button/Button.jsx
```diff
@@ -1,13 +1,42 @@
 import MaterialButton from '@material-ui/core/Button'
+import PropTypes from 'prop-types'
 import React from 'react'
 
-/**
- * Overrides the Material Button component with some custom styles.
- */
+/** The `<Button>` component represents a clickable button. */
```
nullobject: If we use the docgen syntax (e.g. `/** lorem ipsum */`), then the class comment and prop descriptions will be automatically included in the "Overview" section.

nickbrowne:
Really nice changes, it'll be great to not have to look at the source to work out how to use things!

PR #10, opened 2018-12-14
TC-434 Add legacy theme

nullobject:
This PR adds a `legacy` theme to represent the legacy design of our applications. Components configured with this theme should visually gel with our legacy stylesheets. It will be used by default.

I renamed the original theme to the `future` theme. Applications can opt in to this theme, and it will used for future design work of our component library.

It would be nice to be able to switch between these themes in our styleguide, but that can come in a later PR.

screen shot 2018-12-14 at 2 28 56 pm

## How to test

Fire up the storybook with `make storybook`.


nullobject:
@ZoeJazz & @markcipolla: I just plucked a couple of colours from our TC stylesheets for the `primary`, `secondary`, and `error` colours. Let me know if there are more appropriate values to use.

ZoeJazz:
This is fantastic!! Thanks Josh

src/ComponentOverview.jsx
```diff
@@ -1,6 +1,6 @@
 import Marked from 'storybook-readme/components/Marked'
 import React from 'react'
-import { get } from 'lodash'
+import { get } from 'lodash-es'
```
tiagoamaro: omg! I thought `lodash` had modules by default! 😱 TIL

PR #12, opened 2019-01-23
Automatic external dependency configuration for rollup

nickbrowne:
Replaces our manually curated list, which was [causing hard to diagnose problems](https://github.com/conversation/tc/pull/8753) with a plugin that uses our `package.json` to automatically populate the list of external dependencies:

https://www.npmjs.com/package/rollup-plugin-auto-external

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3664,6 +3664,15 @@
       "integrity": "sha1-hZgoeOIbmOHGZCXgPQF0eI9Wnug=",
       "dev": true
     },
+    "builtins": {
+      "version": "2.0.0",
+      "resolved": "https://registry.npmjs.org/builtins/-/builtins-2.0.0.tgz",
+      "integrity": "sha512-8srrxpDx3a950BHYcbse+xMjupHHECvQYnShkoPz2ZLhTBrk/HQO6nWMh4o4ui8YYp2ourGVYXlGqFm+UYQwmA==",
+      "dev": true,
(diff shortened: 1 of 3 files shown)


tiagoamaro:
Storybook and tests are working fine! LGTM and awesome find @nickbrowne 👏

jameshill:
@nickbrowne ha so just to complete my mental model...

lodash should have been included in the externals, it wasn't.. instead of manually keeping it up to date, we can use autoExternal to keep on track of things?

nickbrowne:
That's the one remaining mystery I didn't get to the bottom of, my understanding of webpack and babel is that it should have still transpiled lodash when it was included in the bundle by rollup, but it wasn't for some reason. 

What I changed was basically instructing rollup to leave imports for external libraries intact, so that rather than rollup resolving them, it leaves it up to webpack. Doing that let babel work as intended.

My only guess is that it was a bad interaction between rollup and webpack somehow. I want to chat to Josh about why we're using rollup at all, because the recommendations I could find for frontend libraries seems to be to not use it at all, leaving everything to webpack.

https://medium.com/@kelin2025/so-you-wanna-use-es6-modules-714f48b3a953#fb9f

> When you create a library for modern frontend only
>
> Use ES6 modules and don’t care about. Don’t build it. Webpack will do all work for you.
>
> Just besure that all external dependencies are included to your package.json.

PR #13, opened 2019-01-24
Make imports more specific to avoid circular dependencies

markcipolla:
@nickbrowne and I were looking at how we can fix the spec speeds. Please enjoy 🙏


nullobject:
Big win, thanks ❤️

PR #19, opened 2019-02-06
[TC-567] Add Material-UI Pickers

tiagoamaro:
This PR adds the Material-UI Pickers's `DateTimePicker` component to `@theconversation/ui`.

### Additions

There were a couple of design choices that should be discussed while adding the "Material-UI Pickers" library as a dependency to this project:

- It requires a date management library to parse its dates. Choosing a date management library should be responsibility of the project where `@theconversation/ui` is used, so no explicit date management library was chosen to be added as a dependency.
    - `moment` and `@date-io/moment` were chosen as a `devDependency` so the storybook example works
- All props maps directly to the `material-ui-pickers`'s `DateTimePicker` component, so, instead of recreating their documentation, I found more constructive pointing the reader to their official docs

### Screenshots

image


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@theconversation/ui",
-  "version": "1.8.2",
+  "version": "1.9.0",
```
nullobject: My preference here is that we don't change the version in our feature PRs. We should have a separate PR which bumps the version and updates the changelog. The reason being is that we could potentially have conflicts if multiple people are building different features and they are bumping the version. What do you think?

tiagoamaro: Agreed. Will remove!

tiagoamaro:
@nullobject, just added all three pickers to this PR, adding them with the following folder structure:

```
- src/pickers/stories/overview.jsx
- src/pickers/stories/usage.jsx

- src/pickers/DatePicker
- src/pickers/DateTimePicker
- src/pickers/TimePicker
```

The storybook menu now look like this:

image

Each picker still has its own component overview. The final "Usage" story is where all date and time pickers can be seen in action.

PR #21, opened 2019-02-07
Add ChipInput for TC-48

vertis:
TC-48 is a rework of how institution emails are entered. Currently, they
are comma separated and there is no real validation that this is
correctly entered by the users.

To improve UX with minimal changes under the hood I've introduced a new
type of input called EmailArrayField that supports multiple elements
and returns an array under the hood (that we can then serialize however
we like).

The majority of the heavy lifting is done by an open source library
called 'material-ui-chip-input'. The work done to bring it in is
around state handling, sorting and validation.

After discussions with Zoe, validation is positive (light blue) and
unvalidated entries are simply grey.

There are two new exposed components ChipArrayField and EmailArrayField.
ChipArrayField is the bulk of the styling and state logic, but I've
kept it generic to allow for use in other fields with similar needs.

EmailArrayField contains simple validation logic for emails and
alphabetical sorting as requested by @markcipolla (who does a lot of
support tickets around this field)


src/ChipArrayField/ChipArrayField.jsx
```diff
@@ -0,0 +1,209 @@
...
+      padding: '10px 12px'
+    },
+    chip: {
+      margin: '8px'
+    },
+    label: {
+      top: -12
+    }
+  }
+}
```
tiagoamaro: I see that the border below the label input loose its colour. Is this the intended behaviour?

border-without-color

src/ChipArrayField/ChipArrayField.jsx
```diff
@@ -0,0 +1,209 @@
...
+  sortEntries (entries) {
+    if (this.props.onSortEntry) {
+      return entries.concat().sort(this.props.onSortEntry)
+    }
+    return entries.concat().sort((a, b) => {
+      const valA = a.value.toLowerCase()
+      const valB = b.value.toLowerCase()
+      return (valA < valB) ? -1 : (valA > valB) ? 1 : 0
+    })
+  }
```
tiagoamaro: `entries` prop is an Array. Why the `concat()` empty call?

src/EmailArrayField/EmailArrayField.jsx
```diff
@@ -0,0 +1,75 @@
...
+  helperText: PropTypes.string
+}
+
+EmailArrayField.defaultProps = {
+  disabled: false,
+  entries: [],
+  sorted: false
+}
+
+export default EmailArrayField
```
tiagoamaro: It seems that `EmailArrayField` is a custom implementation of `ChipArrayField`. This could be a story of the `ChipArrayField` instead of being a component itself.

src/ChipArrayField/ChipArrayField.jsx
```diff
@@ -0,0 +1,209 @@
...
+
+  /**
+   * Whether the array should be sorted after each addition
+   */
+  sorted: PropTypes.bool,
+
+  /**
+   * The callback function called to sort the entries.
+   */
+  onSortEntry: PropTypes.func,
```
tiagoamaro: Could detail this docs explaining that it's possible to customize how you sort methods by using this `onSortEntry` callback, such as the `onValidateEntry` docs.

vertis:
@nullobject After watching "A Branch in Time (a story about revision histories)" from RubyConf, I'll clean the commits up and provide a bit more WHY in the description in the morning.

src/ChipArrayField/ChipArrayField.jsx
```diff
@@ -0,0 +1,209 @@
...
+    return entries.map((e) => this.validateEntry(e))
+  }
+
+  validateEntry (e) {
+    let isValid = true
+    if (this.props.onValidateEntry) {
+      isValid = this.props.onValidateEntry(e)
+    }
+    return { value: e, valid: isValid }
+  }
```
tiagoamaro: If you use an always true function as a default prop, there's no need to verify if the prop was configured:

```js
// ... class

  validateEntry (element) {
    let isValid = this.props.onValidateEntry(element);
    return { value: element, valid: isValid }
  }

// ... prop types

ChipArrayField.defaultProps = {
  onValidateEntry: (_element) => true,
  // ... other props
}
```

src/index.js
```diff
@@ -31,4 +31,5 @@ export { default as DialogTitle } from './dialog/DialogTitle'
 export { default as Dropdown } from './Dropdown'
 export { default as Swatch } from './colours/Swatch'
 export { default as TextField } from './TextField'
+export { default as EmailArrayField } from './EmailArrayField'
```
tiagoamaro: Just saw that only the `EmailArrayField` is being exported, while the `ChipArrayField` is not, but both have stories on the storybook.

src/ChipArrayField/ChipArrayField.jsx
vertis: Nope, styling this has been a bit more of a pain than expected.

src/ChipArrayField/ChipArrayField.jsx
vertis: Concat copies the array so that the sort doesn't mutate the existing array.

src/index.js
vertis: Oh, they both should be, that's an oversight on my part.

src/ChipArrayField/ChipArrayField.jsx
vertis: Oh nice, I'll change that.

src/EmailArrayField/EmailArrayField.jsx
vertis: I was trying to keep ChipArrayField free of any email specific logic. I'm open to suggestions though.

src/ChipArrayField/ChipArrayField.jsx
vertis: After digging into this further, it's not something that can be styled through material-ui-chip-input, the solution is to put a pull request on that project, which will take time.

Depends how important we think it is.

vertis:
Okay, I've 'fixup'ed the commits into one commit and written a much more detailed commit message (I've reproduced it in the initial comment above).

I've also included the majority of suggested changes by @tiagoamaro.

screen shot 2019-02-13 at 6 39 19 pm

src/ChipArrayField/ChipArrayField.jsx
tiagoamaro: Since it's too complicated for a small stylistic choice, going to mark this as resolved!

src/ChipArrayField/ChipArrayField.jsx
tiagoamaro: Always forget that there are more than X ways of duplicating arrays on JS.

src/EmailArrayField/EmailArrayField.jsx
tiagoamaro: I think the `EmailArrayField` is a very specific implementation of the `ChipArrayField`. Don't think it needs to be implemented at `@theconversation/ui`.

markcipolla:
The ChipArrayField doesn't seem to wrap when many values are added, like picture below:

screen shot 2019-02-15 at 11 35 49 am

Many institutions have quite a few email addresses in the article publication notification list, so IMO it should wrap to not push off the screen.

markcipolla:
The disabled field doesn't appear disabled, though it's un-interactable.

screen shot 2019-02-15 at 11 38 03 am

Typically, disabled fields are gently greyed out, informing the user without needing to click / try to interact with it to know it's disabled.

vertis:
Based off some of the above feedback and the inability to make the material-ui-chip-input library handle some of the needed changes I've started moving away from this dependency and implementing directly.

This is still a WIP, so I've updated the title to reflect that.

I will build stories/tests for each of the issues you've highlighted above @markcipolla.

@nullobject I will ensure the styling is consistent with the `TextField` component.

markcipolla:
👏 @vertis 
This is v good, mate.

> Many institutions have quite a few email addresses in the article publication notification list, so IMO it should wrap to not push off the screen.

Confirmed it's wrapping as expected 

> The disabled field doesn't appear disabled, though it's un-interactable.

The background now shows that it appears disabled.

There's a bunch of design tweaks we could make, but we can and should do them in follow-up PRs. This is looking great. LGTM 👍

vertis:
I think this is ready to merge, what do I need to do around bumping version numbers etc.

nullobject:
> I think this is ready to merge, what do I need to do around bumping version numbers etc.

Don't worry about bumping the version in this PR.

nullobject:
Looks great. I'm going to merge this, as it's currently working nicely.

We can iterate on it as we need to.

ZoeJazz:
Nice one @vertis :) **This is a note for a separate PR.**

I’d describe this element as a variant of a button so I think there’s an opportunity to follow the same logic as the [main button styles](http://styleguide.theconversation.com/?selectedKind=Buttons&selectedStory=Overview&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel) as well as improve contrast in the blue chip input.

**Recommended update:
Disabled = Outlined button style
Default = Default/contained button style**

Just trying to achieve consistency across the components so we are following the same logic pattern. I think this should allow for an adaptive style that we can easily iterate on later. 

screen shot 2019-02-27 at 8 47 28 am

PR #24, opened 2019-02-19
[TC-604] Use plain ES modules as library main file (drop rollup)

tiagoamaro:
This PR aims to simplify `@theconversation/ui` distribution, by removing `rollup` and using a plain ES with module version to its distribution.

### Why?

Currently, `@theconversation/ui` is using `rollup` to create its distributions, which is bundling `@material-ui/core` to its distribution files, even with `rollup` auto external plugin. This is making other projects to mistakenly import twice several `@material-ui/core` libraries.

Example: if this library's `DateTimePicker` is used inside a `Dialog`, it'll create a `@material-ui/core/Dialog` within a `@theconversation/ui/Dialog`. Both will have their own focus/trigger events to close them, but each library will have its own implementation, making the distributed version call MUI core library and result in a infinite loop recursive call.

### How?

To avoid this issue, this PR proposes using plain ES modules as `@theconversation/ui` entrypoint (`dist/index.js`), leaving the decision on how it's going to be bundled to projects that'll import this library.

### Only distributing JS files

Since this library contains several React components written in JSX, we first want to convert them to plain JS. This is solved on the distribution process.

Instead of callling `rollup`, we process all `src` files through `@babel/cli` and push its results to the `dist` folder. This way, the distribution process is kept the same, but we now have `@theconversation/ui` source available to be used by other projects.

### How to test it?

Since `npm` can be quite tricky when handling local modules, I've created a temporary branch called `material-ui-as-external-dependency-WITH-DEPS`, where I've pushed the distribution version to github. To use it, simply install through the command: `npm install "conversation/ui#material-ui-as-external-dependency-WITH-DEPS" --save-dev`.


package.json
```diff
@@ -25,35 +24,28 @@
   },
   "dependencies": {
     "@material-ui/core": "^3.9.1",
-    "@material-ui/icons": "^3.0.2",
     "lodash-es": "^4.17.11",
-    "material-ui-pickers": "^2.1.2"
+    "material-ui-pickers": "^2.1.2",
+    "react-jss": "^8.6.1"
```
tiagoamaro: `react-jss` was being bundled together when using `rollup`. With this approach, we make it a requirement, but other bundlers will be able to tree-shake it away.

tiagoamaro:
Also, keeping in mind we would be serving an "untranspiled" package, which will require some `jest` configuration, according to its docs:

> Sometimes it happens, ..., that 3rd party modules are published as untranspiled. Since all files inside node_modules are not transformed by default, Jest will not understand the code in these modules, resulting in syntax errors. - https://jestjs.io/docs/en/configuration#transformignorepatterns-array-string

package.json
nullobject: Fair enough, we have a hard dependency on it in `ThemeProvider.jsx`.

nullobject:
@nickbrowne Do you want to cast your 👀 over this before I merge it?

nickbrowne:
Sure, I'll try it out shortly

nickbrowne:
I just tried this branch on https://github.com/conversation/ssr-frontend-example and ran into a couple of small issues to do with leaving some of the ES6 intact.

After changing the babel config a bit to transpile ES6 imports/exports to CJS and swapping `lodash-es` for regular `lodash`, it works just fine again.

@nullobject those changes all good with you?

nullobject:
@nickbrowne Yay for regular lodash 🎉 

I'll just have a play around locally to check it out.

nickbrowne:
Cool, if you're happy feel free to merge.

nullobject:
For the record, there was a bit of a discussion in Slack about the choice to switch back to CJS:

```
josh [2:44 PM]
@nickbrowne @tiagoamaro Does compiling UI to CJS affect tree shaking in any way? i.e. is it better to leave things as ES6?

nickbrowne [2:45 PM]
it doesnt in our case, tree shaking requirements are very strict, you have to actually specify a certain property in your `package.json` for webpack 4 to attempt tree shaking, which basically tells it your library is guaranteed side effect free (edited) 
since we didn't have that before, I believe effectively nothing has changed

tiagoamaro [2:45 PM]
Also, https://github.com/indutny/webpack-common-shake
indutny/webpack-common-shake
CommonJS Tree Shaker plugin for WebPack
Stars
803
Language
JavaScript
indutny/webpack-common-shakeJun 29th, 2017Added by GitHub
Since there’s linting happening on `@theconversation/ui` to guarantee no abandoned module usage, don’t think it’ll be a problem

nickbrowne [2:47 PM]
https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free
webpack
Tree Shaking | webpack
webpack is a module bundler. Its main purpose is to bundle JavaScript files for usage in a browser, yet it is also capable of transforming, bundling, or packaging just about any resource or asset.

josh [2:48 PM]
that would be pretty easy to do, UI shouldn't have side effects

nickbrowne [2:49 PM]
yeah, we should probably have it, but if one of the dependencies doesn't have that option, I think it might "poison" that whole branch of the tree

josh [2:52 PM]
@nickbrowne what was the issue you ran into with SSR?
(sorry just trying to understand everything before i merge)

nickbrowne [2:53 PM]
node, at least LTS 8, can't understand the es6 imports and exports
it was more a convenience thing, where you need less configuration when actually trying to use the library

josh [2:56 PM]
Right, so if we were to leave ES6 in UI then the SSR server side code would need to be transpiled?

nickbrowne [2:57 PM]
yeah, well - the server code is already transpiled (using es6 internally), but just not anything in node_modules
which is why I had to switch lodash-es too

josh [2:59 PM]
Cool, let's roll with CJS and if some pressing need to package it as ES6 arises, then we can revisit this decision.

nickbrowne [2:59 PM]
yeah totally, I feel like we're still learning about what the best way to do all this is :slightly_smiling_face:
right now I'm mostly concerned about trying to minimise config so things mostly Just Work as much as possible

josh [3:03 PM]
i agree, developer happiness is the best thing to optimise for
Message Input
```

PR #25, opened 2019-02-26
[RFC] Use Noto Sans font-family for body2

tiagoamaro:
I've notice this before, but confirmed today with @ZoeJazz that the `body2` variant typography seemed "not quite fit" for its use cases. I confirmed with her that "Noto Sans" should be the font to be used on any type of "labeled information", while "Libre Baskerville" will be used for content reading due to being a serif font.

### Before

----

#### Switch

image

----

#### Typography Variants

image

### After

----

#### Switch

image

----

#### Typography Variants

image

### DIFF
--- a/src/styles/themes/legacy.js
+++ b/src/styles/themes/legacy.js
@@ -25,9 +25,9 @@ export default memoize(() => createMuiTheme({
     },
 
     body2: {
-      fontFamily: 'Libre Baskerville',
       fontSize: '1rem',
       lineHeight: '1.5'
+      fontFamily: 'Noto Sans'
     },
 
     button: {
--- a/src/styles/themes/legacy.js
+++ b/src/styles/themes/legacy.js
@@ -25,8 +25,6 @@ export default memoize(() => createMuiTheme({
     },
 
     body2: {
-      fontSize: '1rem',
-      lineHeight: '1.5'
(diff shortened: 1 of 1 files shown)


nullobject:
Much better.

My feels are that we should probably use Libre Baskerville for `body1`, Otherwise `body1` and `body2` will both be sans serif. Thoughts?

tiagoamaro:
I think what MUI considers as `body1` differs from what we see as "body of text". Example, Baskerville would be used for reading content, such as articles' excerpts.

If I change `body1` styling to be "Baskerville", this will happen:

- Before

image

- After

image

----

It seems that MUI uses the `body1` in different places where not necessarily the typography is a "body", such as the card subheader.

I imagine the serifed font would only be used if we explicitly specified it as "body of content". Is it possible to add a custom typography element?

nullobject:
@tiagoamaro Fair enough. I notice that the default MUI theme has the same font (Roboto) for all styles.

Are we happy running Montserrat just for our `body1` style?

tiagoamaro:
It would be interesting if MUI specified where their typography hierarchy is applied, but as you mentioned, it seems they use a single font and that's it. Maybe we could leave the "Montserrat" as the general one, and "Noto Sans" for labels and "utility" labels and such (leaving the config as it is on this PR).

PR #27, opened 2019-03-18
[TC-688] Add events to TextField and Dropdown

markcipolla:
For TC Donations, we use `onBlur` events to do some Google tracking. For example, as a donor progresses through the form, we track a number of events as to which fields they completed, or left blank, etc...

However, we only explicitly passed along `onChange`. I've added `onBlur` and `onFocus` to the `TextField` and `Dropdown` components, plus added specs and demonstrations in the storybook.

### DIFF
--- a/src/Dropdown/stories/events.jsx
+++ b/src/Dropdown/stories/events.jsx
@@ -1,0 +1,65 @@
+import React from 'react'
+import { action } from '@storybook/addon-actions'
+import { withDocs } from 'storybook-readme'
+
+import { GridLayout } from '../../util'
+import { Dropdown, MenuItem } from '../../index'
+
+const md = `
+# States
+
+The \`<Dropdown>\` component accepts multiple events:
+
+* \`onChange\`: When the field's value changes
+* \`onBlur\`: When the field loses focus
+* \`onFocus\`: When the field gains focus
+
+~~~js
+<Dropdown onChange={alert('change')} />
+<Dropdown onBlur={alert('blur')} />
+<Dropdown onFocus={alert('focus')} />
+~~~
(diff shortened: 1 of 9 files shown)


src/Dropdown/stories/events.jsx
```diff
@@ -0,0 +1,65 @@
...
+      >
+        <MenuItem value='1'>One</MenuItem>
+        <MenuItem value='2'>Two</MenuItem>
+        <MenuItem value='3'>Three</MenuItem>
+      </Dropdown>
+    )
+  }
+}
+
+export default withDocs(md, () =>
```
nullobject: What about just having only one example dropdown wired up to all of the event handlers? That way you don't have to click into the correct dropdown to see how generated events.

src/TextField/stories/events.jsx
```diff
@@ -0,0 +1,46 @@
...
+<TextField onFocus={alert('focus')} />
+~~~
+
+## Example
+
+<!-- STORY -->
+`
+
+export default withDocs(md, () =>
+  <GridLayout>
```
nullobject: What about just having only one example text field wired up to all of the event handlers? That way you don't have to click into the correct text field to see how generated events.

src/Dropdown/stories/events.jsx
markcipolla: Done!

src/TextField/stories/events.jsx
markcipolla: Done!

markcipolla:
Added a note in the `CHANGELOG` as well.

PR #29, opened 2019-03-25
[TC-736] Add MenuItem to TCUI, instead of calling @material-ui directly.

markcipolla:
REFS https://github.com/conversation/tc-donations/pull/702#discussion_r267878803

@tiagoamaro _3 days ago_
I think we should avoid referencing @material-ui as much as possible, since these should be references of @theconversation/ui.

@tiagoamarotiagoamaro _3 days ago_
Also, if the MenuItem MUI delegation does not exist, a new @theconversation/ui should be released with it :)

@markcipollamarkcipolla _3 days ago_ 
Hmmm.... yeah, good point. I'll add it to the TCUI lib.

### DIFF
--- a/src/MenuItem/MenuItem.jsx
+++ b/src/MenuItem/MenuItem.jsx
@@ -1,0 +1,57 @@
+import MaterialMenuItem from '@material-ui/core/MenuItem'
+import PropTypes from 'prop-types'
+import React from 'react'
+
+/**
+ * The `<MenuItem>` component functionally replaces what would be `option`
+ * elements in a `select` field.
+ #
+ * MenuItems are passed to `Dropdown` components as children.
+ *
+ * ```js
+ * import { Dropdown, MenuItem } from '@theconversation/ui'
+ *
+ * <Dropdown
+ *   label="Country"
+ *   value="AU"
+ * >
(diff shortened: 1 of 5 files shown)


tiagoamaro:
@markcipolla, I just noticed that the `MenuItem` is already exported by `@theconversation/ui`: https://github.com/conversation/ui/blob/v1.13.0/src/index.js#L17. Didn't see this before! Sorry! :bowing_man: 

It should be just a matter of importing it:

```js
import { MenuItem } from "@theconversation/ui";
```

----

While we are at it, this PR might serve as a discussion about how should we document these "delegated" MUI components, as one might not know they are already available through the `@theconversation/ui` package (unless trying to use an editor/IDE auto-completion or checking the distribution `index.js`).

What do you think @nullobject @markcipolla? How can we make explicit for us on which of these [delegated MUI components](https://github.com/conversation/ui/blob/v1.13.0/src/index.js#L10-L21) are available through `@theconversation/ui`?

markcipolla:
> Didn't see this before! Sorry!

Neither did I! Hmmm. Well, I'll close this PR, as it's unnecessary. But yes, a great question.

PR #30, opened 2019-03-26
[TC-746] Add ToggleButton and ToggleButtonGroup from @material-ui/lab

markcipolla:
@tiagoamaro and @nullobject - I'm thinking of using https://material-ui.com/lab/api/toggle-button-group and the https://material-ui.com/lab/api/toggle-button

Is this how we want to include new components? Do we want to wrap them in our own components?


nullobject:
It's a shame the toggle button is in the lab package. I guess the danger is that they radically change the API of the component, or remove it completely. I guess if that happens we could just add our own toggle button.

Do you have any feels about this?

markcipolla:
Just unsure if we want to roll our own or use theirs; and how much to tie our implementation to theirs. I'm very happy to be led by what you think is best.

nullobject:
Sure, I think let's roll with their version until there's a real issue 👍

PR #34, opened 2019-04-01
Add id, onBlur, and onFocus props to Autocomplete component

markcipolla:
Seems Downshift generates an id (like `downshift-[INTEGER]-input`), and hijacks the events, to make it work the way it wants (to display an autocomplete).

I've explicitly set the id and onBlur so it uses a given prop, or the generated downshift ones.

@nullobject and @tiagoamaro, I'd love your expertise on this.

### DIFF
--- a/src/Autocomplete/stories/events.jsx
+++ b/src/Autocomplete/stories/events.jsx
@@ -1,0 +1,39 @@
+import React from 'react'
+import { action } from '@storybook/addon-actions'
+import { withDocs } from 'storybook-readme'
+
+import { GridLayout } from '../../util'
+import { Autocomplete } from '../../index'
+import countries from './countries'
+
+const md = `
+# Events
+
+The \`<Autocomplete>\` component accepts multiple events:
+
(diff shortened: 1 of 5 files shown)


nullobject:
@markcipolla I made some tweaks. It's better to use the [Downshift API](https://github.com/downshift-js/downshift#getinputprops) to override these props.

src/Autocomplete/stories/events.jsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react'
+import { action } from '@storybook/addon-actions'
+import { withDocs } from 'storybook-readme'
+
+import { GridLayout } from '../../util'
+import { Autocomplete } from '../../index'
+import countries from './countries'
+
+const md = `
+# Events
```
markcipolla: 😗👌

nullobject: Just keeping it consistent with the other components.

markcipolla: Love it.

PR #35, opened 2019-04-01
[TC-805] Allow Textfield to pass autoComplete prop

markcipolla:
This PR adds the `autoComplete` prop to `TextField`, allowing us to set it both `on` and `off`, but also to more specific browser-helper strings (read more at https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofilling-form-controls:-the-autocomplete-attribute)

### DIFF
--- a/src/TextField/TextField.jsx
+++ b/src/TextField/TextField.jsx
@@ -63,7 +63,6 @@ export const TextField = ({
   ...other
 }) => {
   const helperTextId = helperText && id ? `${id}-helper-text` : undefined
-
   return (
     <FormControl aria-describedby={helperTextId} disabled={disabled} error={error} fullWidth={fullWidth}>
       {label && (
@@ -84,6 +83,11 @@ export const TextField = ({
 }
 
 TextField.propTypes = {
+  /**
+   * Overrides the styles applied to the component.
+   */
+  autoComplete: PropTypes.string,
+
   /**
    * Overrides the styles applied to the component.
    */
@@ -158,6 +162,7 @@ TextField.propTypes = {
 }
 
 TextField.defaultProps = {
+  autoComplete: 'off',
   disabled: false,
   error: false,
   fullWidth: false,
--- a/src/TextField/TextField.test.jsx
+++ b/src/TextField/TextField.test.jsx
@@ -15,6 +15,15 @@ describe('<TextField />', () => {
     expect(wrapper.props().value).toBe('foo')
   })
 
+  describe('with an autoComplete setting', () => {
+    it('sets the autocomplete attr', () => {
+      const wrapper = shallow(
+        <TextField autoComplete='postal-code' />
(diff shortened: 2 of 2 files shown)


markcipolla:
> Even at this early stage, my feels are that adding this to the `Autocomplete` field doesn't make a lot of sense. When would you ever want to turn on native autocomplete for a Downshift autocomplete component?
> 
> For the `TextField` component, this definitely makes sense.

My spider sense defs agrees. Like, ideally, I'd _love_ the donations form to autocomplete the country if the browser can... but downshift seems to allow it:

https://github.com/downshift-js/downshift/issues/497
https://github.com/downshift-js/downshift/issues/457

```
<Downshift>
  {({
    ...
    getInputProps,
    ...
  }) => (
    <input
      {...getInputProps({
        ...
        autoComplete: 'off',
        ...
      })}
    />
  )}
</Downshift>
```

markcipolla:
@nullobject How about we pivot this PR to add it for the `TextField` components for now, and we can read more and investigate the ramifications of changing the autocomplete prop on `Autocomplete` and Downshift?

src/TextField/TextField.jsx
```diff
@@ -84,6 +83,11 @@ export const TextField = ({
 }
 
 TextField.propTypes = {
+  /**
+   * Overrides the styles applied to the component.
+   */
+  autoComplete: PropTypes.string,
```
nullobject: My preference would be to turn this into a boolean, rather than a string. The text field component can take care of translating this into a `on` or `off` value.

markcipolla: Ah! 

It's intentionally a string: you can give helper information to the browsers by setting different values:

https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fe-autocomplete-nickname

nullobject:
> @nullobject How about we pivot this PR to add it for the TextField components for now, and we can read more and investigate the ramifications of changing the autocomplete prop on Autocomplete and Downshift?

Yep, totally agree.

src/TextField/TextField.jsx
markcipolla: > Yep, totally agree.

I've rebased out the `Autocomplete` component commit.

nullobject: TIL

markcipolla: For 99% of the cases, a boolean would've been fine. I'm thinking there might be specific situations (like adding a country selector) where we might have to be more specific with it.

This bit made me 🤯:

The "on" keyword indicates that the user agent is allowed to provide the user with autocompletion values, but does not provide any further information about what kind of data the user might be expected to enter. **User agents would have to use heuristics to decide what autocompletion values to suggest.**

markcipolla:
Well, no need for this PR then :)

PR #36, opened 2019-04-01
Allow Autocomplete to pass the name prop to the underlying Textfield

markcipolla:
This PR allows the Autocomplete to pass the input field's `name` prop to the underlying TextField.

TC Donations uses the name prop (ala `donation["country"]`) to validate the fields, and send the response to the server. Without it, the `country` attr on the `Subscription` is `nil`.

### DIFF
--- a/src/Autocomplete/Autocomplete.test.jsx
+++ b/src/Autocomplete/Autocomplete.test.jsx
@@ -44,6 +44,22 @@ describe('<Autocomplete />', () => {
     })
   })
 
+  describe('with a name', () => {
+    it('sets the input field name', () => {
+      const wrapper = mount(
+        <Autocomplete name='name' />
+      ).find(TextField)
+      expect(wrapper.props().name).toBe('name')
+    })
+
+    it('does not add a name field', () => {
+      const wrapper = mount(
+        <Autocomplete />
+      ).find(TextField)
+      expect(wrapper.props().name).toBe(undefined)
+    })
+  })
+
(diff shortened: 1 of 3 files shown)


markcipolla:
> So donations is still doing a form submit at the end of the day?

Yep.

> My feeling is that we should be moving away from using the DOM for form stuff (i.e. we should build/validate an object client-side, and submit it to the server).

Agreed, but minimal changes at one time. Moving away from a form submission is a good goal to work towards.

> If we want to continue down this path, then won't we need to be able to supply the `name` prop to the all of the other input components too?

We already do; this is to make the `Autocomplete` pass it to the `Textfield` underneath.

nullobject:
@markcipolla Of course, the good old `...other` prop 😜

src/Autocomplete/Autocomplete.test.jsx
```diff
@@ -44,6 +44,23 @@ describe('<Autocomplete />', () => {
...
+        <Autocomplete name='name' />
+      ).find(TextField)
+      expect(wrapper.props().name).toBe('name')
+    })
+
+    it('does not add a name field', () => {
+      const wrapper = mount(
+        <Autocomplete />
+      ).find(TextField)
+      console.log(wrapper.props().name)
```
nullobject: Oops

markcipolla: Removed and force pushed.

giphy (5)

PR #38, opened 2019-04-03
Add Donation Dialog (and supporting components)

nickbrowne:
Building on some of the recent changes to themes/palettes this adds the Donation Dialog (aka popup) and various supporting components.

Screenshot at 2019-04-10 17-49-36

Had to add a little extra config to jest so it doesn't blow up when importing an image (default for `Avatar`), but now works great.

See the new storybook entries for more details on the implementation along with the working examples.

Still running into a few clunky areas to do with theming, but we can continue refining that outside of this branch (I've left some comments where there are some remaining issues).


nickbrowne:
Now good to review!

src/Avatar/Avatar.jsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import MaterialAvatar from '@material-ui/core/Avatar'
+import withStyles from '@material-ui/core/styles/withStyles'
+import defaultImage from './default.jpg'
```
tiagoamaro: Default avatar has `1.251 bytes` :clap:

src/Avatar/Avatar.jsx
```diff
@@ -0,0 +1,48 @@
...
+ * import { Avatar } from '@theconversation/ui'
+ *
+ * <Avatar src={YOUR_URL} />
+ * ```
+ */
+export const Avatar = ({ src, size, classes }) => {
+  return (
+    <MaterialAvatar src={src} classes={classes} style={{ width: size, height: size }} />
+  )
+}
```
tiagoamaro: Is it worth it adding `...other` to the `Avatar` props, so any other MUI can be passed down to the [`MaterialAvatar` API](https://material-ui.com/api/avatar/)?

nullobject:
Rebased and force pushed.

jameshill:
🎉 🚀 

amazing work @nickbrowne @tiagoamaro @nullobject

PR #39, opened 2019-04-04
[TC-822] Allow specifying the autoComplete prop as a boolean or string

nullobject:
You can now specify the `autoComplete` prop as either a boolean or string value for the `TextField` component.

This little bit of sugar should it a touch nicer to use. For example, you don't need to use `autoComplete='on'` to enable autocomplete anymore:

    <TextField autoComplete />

or if you want anything else:

    <TextField autoComplete='email' />

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,8 @@
-##1.18.0 (2019-04-02)
+## Unreleased
+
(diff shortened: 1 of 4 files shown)


markcipolla:
LGTM 👍

PR #42, opened 2019-04-08
Unify colours and typography under single theme

ZoeJazz:
This change is an adaptation on the existing 'future' colours in storybook. Though it comes with some issues that I'd love your help with @nullobject / @nickbrowne

Proposed colours and naming convention:
New colours

Working with material's grid for colours, there are now a lot of blanks
New colours in Storybook

Lastly, this change messes up the 'legacy' stylesheet so there's a question around whether we should retain this or not.


src/styles/stories/colours.jsx
```diff
@@ -3,15 +3,22 @@ import React from 'react'
...
+const COLOURS = ['neutral', 'core', 'accent', 'positive', 'aware']
 
 export default () => (
   <GridLayout>
     {COLOURS.map((colour, index) =>
       <div key={index}>
-        <Swatch colour={colour} />
-        <Swatch colour={colour} variant='dark' />
-        <Swatch colour={colour} variant='light' />
+        {/* <Swatch colour={colour} /> */}
```
nullobject: Was this comment just temporary?

tiagoamaro:
This looks very promising! I can see there are are a couple of changes required to the `Swatch` component:

- `colour` `PropType` needs to accept the future theme naming

https://github.com/conversation/ui/blob/3d580734b00c13311ff4c17b670a03653c4b4be0/src/colours/Swatch.jsx#L54

- `COLOURS.map` `future` theme iterator must but backward compatible with the `legacy` theme, otherwise storybook will :boom: 

image

ZoeJazz:
> I haven't had a chance to play around to see how we can make it work with the built-in palette intentions, but I'm guessing it's possible.
+ 
> COLOURS.map future theme iterator must but backward compatible with the legacy theme

Yeah totally! I started playing around with a solution for both but it got me wondering whether we even need the legacy theme in this space. What do you think?

@nullobject @tiagoamaro @nickbrowne

nickbrowne:
Yesterday I was playing with making the colour palette documentation more dynamic so it just looks up the colours available, but it quickly became a lot more complicated than expected.

I'm thinking for now, it might be better to remove the multiple theme support just so we can keep moving... or we just plonk in some placeholders or duplicate the colours across both themes for now. Any thoughts?

nullobject:
@ZoeJazz @nickbrowne I'd be happy to shelve the multiple theme stuff for now.

nickbrowne:
Cool, I think it makes sense. Initially the future theme was going to be a much bigger divergence than what we've currently got, whereas now it's more of an evolution. Okay if I make those changes on this branch? cc @ZoeJazz

ZoeJazz:
> Cool, I think it makes sense. Initially the future theme was going to be a much bigger divergence than what we've currently got, whereas now it's more of an evolution. Okay if I make those changes on this branch? cc @ZoeJazz

yes please :) thank you

nickbrowne:
@ZoeJazz I've made changes to the theme to give us just enough to push on with the donations work in a consistent way. Let's look at this again (colours in particular) once we've got some of those components out the door.

nickbrowne:
@nullobject mind taking another look at this and merging if you're happy? @ZoeJazz is :+1:

nullobject:
> @nullobject mind taking another look at this and merging if you're happy?

Sure can.

nullobject:
One thing I noticed is that the text field error state is missing the red border colour:

Screen Shot 2019-04-09 at 5 52 35 pm

nickbrowne:
Nice spotting, hopefully easy enough to fix.

nickbrowne:
Fixed, error colours match current master again.

nickbrowne:
This is all good for now, but we'll be revisiting it after donations campaign.

PR #44, opened 2019-04-09
[TC-842] Rewrite TextField as a class rather than a function.

markcipolla:
I want to, in an outer component on TC Donations, add a ref to a
TextField, but you can't add refs to functions, only classes.

https://reactjs.org/docs/refs-and-the-dom.html#refs-and-function-components

### DIFF
--- a/src/TextField/TextField.jsx
+++ b/src/TextField/TextField.jsx
@@ -53,42 +53,48 @@ const styles = theme => {
  * />
  * ```
  */
-export const TextField = ({
-  InputProps,
-  autoComplete,
-  disabled,
-  error,
-  fullWidth,
-  helperText,
-  id,
-  label,
-  required,
-  ...other
-}) => {
-  const helperTextId = helperText && id ? `${id}-helper-text` : undefined
+class TextField extends React.Component {
+  render () {
+    const {
+      InputProps,
+      disabled,
+      error,
+      fullWidth,
+      helperText,
+      id,
+      label,
+      required,
+      ...other
+    } = this.props
 
-  // Sugar for setting autoComplete using a boolean value.
-  if (isBoolean(autoComplete)) {
-    autoComplete = autoComplete ? 'on' : 'off'
+    const helperTextId = helperText && id ? `${id}-helper-text` : undefined
+
+    // Sugar for setting autoComplete using a boolean value.
+    let autoComplete = this.props.autoComplete
+    if (isBoolean(autoComplete)) {
+      autoComplete = autoComplete ? 'on' : 'off'
+    }
+
+    return (
+      <FormControl
+        aria-describedby={helperTextId} disabled={disabled} error={error} fullWidth={fullWidth}
+      >
+        {label && (
+          <FormLabel htmlFor={id} required={required}>
+            {label}
+          </FormLabel>
+        )}
+
+        <InputBase {...other} autoComplete={autoComplete} id={id} {...InputProps} />
+
+        {helperText && (
+          <FormHelperText id={helperTextId}>
+            {helperText}
+          </FormHelperText>
+        )}
+      </FormControl>
+    )
   }
-
-  return (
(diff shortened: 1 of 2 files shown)


.ruby-version
```diff
@@ -0,0 +1 @@
+2.6.1
```
benmacleod: Do we need this?

markcipolla: Eeeep, no. I'll remove.

src/TextField/TextField.jsx
```diff
@@ -53,42 +53,48 @@ const styles = theme => {
...
-  }
+class TextField extends React.Component {
+  render () {
+    const {
+      disabled,
+      error,
+      fullWidth,
+      helperText,
+      id,
+      InputProps,
```
benmacleod: Some disagreement over how to sort capital letter IDs here. I think the general convention observed in the project is to sort capital letters first 😬

markcipolla: From their docs:
Screen Shot 2019-04-09 at 3 46 08 pm

markcipolla: WHUT

markcipolla: TIL that capitalised props go first before non-capitalised props.

markcipolla:
If you're happy, pls merge-a-diddlie-dooo!

nullobject:
Hmm, that's a shame. It was nice having everything as functional components.

markcipolla:
> Hmm, that's a shame. It was nice having everything as functional components.

Agreed; since it's only for a `ref`, and not for any `state` we'd get, yeah...

nullobject:
@markcipolla I'm not saying "no", just pausing to think for a moment 😄

markcipolla:
> Where are we using the refs, if you don't mind me asking? Maybe it's better to do away with the need for refs in the dependent project?
> 
> My feels are that _if_ we need to do this for `Text`

TC Donations in `CustomAmountButton`. It uses the ref to give focus and to get to the value of the custom field (the `Textfield`) when clicked by _another_ element (hence not using `event.target.value`).

markcipolla:
> My feels are that _if_ we need to do this for `TextField`, then we should rewrite all the components as classes. Otherwise things are inconsistent.

Up to you. There isn't really a difference from the app _using_ TCUI if a component is a function or a class.

To be honest, I think that not being able to `ref` a function component is a bit silly, but I'm sure there's deep reasons.

markcipolla:
> I'm a bit reluctant to do that however, because I prefer the functional component style. The react project also seems to be going down the functional component route, now that they have added hooks. IMHO they have flagged that the future of react is in functional components.

Fair point. Perhaps it's in a state of flux, and we will be able to `ref` functional components _in the future_... but their docs explicitly state we can't. 😢

nullobject:
Just been reading...

We can certainly use refs with functional components, but we would need to upgrade to react 16.8.0 to use the new hooks API: https://reactjs.org/docs/hooks-reference.html#useref

markcipolla:
> Just been reading...
> 
> We can certainly use refs with functional components, but we would need to upgrade to react 16.8.0 to use the new hooks API: https://reactjs.org/docs/hooks-reference.html#useref

Good find. Directly contradicts their own docs. 
https://reactjs.org/docs/refs-and-the-dom.html#refs-and-function-components

Guess they've moving ruul fast!

nickbrowne:
I'm pretty keen to keep all the components functional, so I vote for using the new hooks API, which looks pretty simple, and has support for SSR.

nullobject:
@markcipolla Hit me up. I'll help you to get tc-donations upgraded to react 16.8.0 so we can use the hooks API.

markcipolla:
Let's pair tomorrow morning 👏👏👏

markcipolla:
No more need for this, as bumping React/ReactDOM to 16.8.x allowed for functional classes to be referenced internally with `refs`

PR #45, opened 2019-04-10
[TC-846] Add new named colours

nullobject:
This PR restores the named colours that @ZoeJazz and @jameshill originally added in #42.

These colours are all captured in separate files, so that we can access them internally or from any of our external projects. The Material UI project is [structured similarly](https://github.com/mui-org/material-ui/blob/next/packages/material-ui/src/colors/index.js).

To use them in another project:

```js
import { colours } from '@theconversation/ui'
colours.accent[200]
```

I also moved the `Swatch` component into its own directory, so that the `colours` directory only contains our colour files.


ZoeJazz:
Nice work @nullobject - I've crossed checked the colours with the sketch file and they're perfecto.

PR #48, opened 2019-04-11
[TC-855] Set the DonationDialog actions bottom margin to 32px

nullobject:
This PR makes an adjustment to the `DonationDialog` actions bottom margin.

By using a 4x multiple of the `theme.spacing.unit` variable (which is 8px by default), we end up with 32px. This means that we can scale this margin with the rest of the spacing, if we change the theme spacing later.

Screen Shot 2019-04-11 at 4 46 10 pm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased
+
+* Set the `DonationDialog` actions bottom margin to 32px
+
 ## 1.19.2 (2019-04-11)
 
 * Set ARIA label on `DonationDialog` close button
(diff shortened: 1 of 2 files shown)


jameshill:
Looks good to me... I've got it running locally.

When @ZoeJazz get's a minutes I'll get her to take a squizz.

jameshill:
@nullobject @ZoeJazz is keen to go with 6x. I've pushed a commit...

Does this work for you? Shouldn't have any other unintended consequences as just scoped to donation dialog.

nullobject:
@jameshill Fine by me, if @ZoeJazz is happy I'm happy 👍

PR #49, opened 2019-04-11
Add article donation banner

nickbrowne:
This adds the article donation banner component. The component itself is really pretty simple, so the value is mostly in the examples. The child node splitting is the only real complicated part, this is mostly due to some unanswered questions to do with our themes and colour palettes that we should revisit.

This is what it looks like:

Screenshot at 2019-04-11 18-32-43

I haven't added the "small layout" detection stuff yet. Mostly so we can get this into TC now. Will start working on that now (for merging in another PR).


jameshill:
sweeeet.

PR #52, opened 2019-04-18
Allow type number for TextField input

jameshill:
Dependency of https://github.com/conversation/tc-donations/pull/729


markcipolla:
👏

PR #55, opened 2019-04-24
TC-780 Add top banners

nickbrowne:
Adds two new components for the donation campaign:

## DonationBanner
Screenshot at 2019-04-24 13-31-46

## ThinDonationBanner
Screenshot at 2019-04-24 13-31-57

### DIFF
--- a/src/promo/DonationBanner.jsx
+++ b/src/promo/DonationBanner.jsx
@@ -1,0 +1,102 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import withStyles from '@material-ui/core/styles/withStyles'
+import Paper from '@material-ui/core/Paper'
+import MaterialIconButton from '@material-ui/core/IconButton'
+import CloseIcon from '@material-ui/icons/Close'
+import Button from '../Button'
+import Typography from '../Typography'
+
+const styles = theme => ({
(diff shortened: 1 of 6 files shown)


nickbrowne:
You may notice that the "x" icon is slightly further in from the designs on zeplin, this matches the position in the `DonationDialog`, and is to give the hover/active animations some breathing room. Has Zoe's approval.

src/promo/DonationBanner.jsx
```diff
@@ -0,0 +1,102 @@
...
+  onClose: PropTypes.func,
+
+  /**
+   * Accessibility text for close button
+   */
+  closeText: PropTypes.string
+}
+
+DonationBanner.defaultProps = {
+  href: 'http://example.com',
```
jameshill: Should we just make the default https://donate.theconversation.com?

jameshill: We didn't make it the default on the `EndOfArticle`, but just thought going forward perhaps we should.

jameshill:
Given that the default is also in `EndOfArticle` i'm voting for changing it in a separate PR.

nickbrowne:
Yeah it's currently `'http://example.com'` for `DonationDialog` as well.

PR #59, opened 2019-04-29
[TC-900] Fix overflow when focused on DropDown components

markcipolla:
We were seeing a faint background colour to the right of the `Dropdown` when `:focus`; the `padding-right` was a smidge too big.

#### Before:
Screen Shot 2019-04-29 at 4 40 28 pm

### After:
Screen Shot 2019-04-29 at 4 40 34 pm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased
+
+* Minor design fix to `Dropdown` to avoid `:focus` overflow of background
+
 ## 2.0.0 (2019-04-29)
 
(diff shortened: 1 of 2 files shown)


nullobject:
@markcipolla Can you chuck something in the CHANGELOG under the unreleased section.

markcipolla:
> @markcipolla Can you chuck something in the CHANGELOG under the unreleased section?

Of course; I keep forgetting the process to changelog management. Done. Thanks for reminding me.

PR #63, opened 2019-05-02
[WIP] Update DonationDialog breakpoints

nullobject:
This PR changes the rendering bahaviour of the `DonationDialog` component.

May-02-2019 10-56-27

On small devices, we want to pin the dialog to the top (rather than vertically centre it) and remove the horizontal margins.


tiagoamaro:
Closing due to inactivity.

PR #67, opened 2019-05-14
Amend pop up margin on mobile to match original design

nullobject:
This PR updates the breakpoints for the `DonationDialog` component.

With these new breakpoints, we have:

- For small devices, a 24px horizontal margin around the dialog. It is anchored to the top of the screen and will grow vertically to fit the the dialog content.
- For medium-large devices, the dialog is vertically and horizontally centred on the page.

If anyone knows an easier way to do this, I am all ears 👂

## How to Test

- Fire up storybook and check out the `DonationDialog` at various sizes.

### Before

Small:

IMG_3813

Large:

localhost_9001_iframe html_id=promos--donationdialog(iPad) (3)

### After

Small:

IMG_3812

Large:

localhost_9001_iframe html_id=promos--donationdialog(iPad) (2)

### DIFF
--- a/src/promo/DonationDialog.jsx
+++ b/src/promo/DonationDialog.jsx
@@ -9,13 +9,40 @@ import DialogActions from '../dialog/DialogActions'
 import Button from '../Button'
 
 const styles = theme => ({
+  container: {
+    // For small-medium devices, render the dialog container normally.
+    [theme.breakpoints.down('sm')]: {
+      display: 'block',
+      overflowY: 'auto',
+      overflowX: 'hidden'
+    },
+
+    // For medium-large devices, render the dialog container in the centre.
+    [theme.breakpoints.up('sm')]: {
+      display: 'flex',
+      justifyContent: 'center',
+      alignItems: 'center'
+    }
+  },
   paper: {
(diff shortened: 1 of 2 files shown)


nullobject:
[[Promos] Amend pop up margin on mobile to match original design](https://trello.com/c/FxgZlq1E/82-bug-amend-promos-margin-to-match-original-design)

src/promo/DonationDialog.jsx
```diff
@@ -79,14 +106,14 @@ export const DonationDialog = ({
     open,
     onClose,
     onEnter: onVisible,
-    scroll: 'body',
```
nickbrowne: Sorry I didn't get a chance to review this before it was merged, but the reason we were using `scroll: 'body'` is to work around this issue:

Screenshot at 2019-05-14 16-52-44

I think it's a much better user experience if they can scroll the whole page to see the dialog contents (which is also the existing behaviour of our "legacy" lightboxes in TC instead of having an internal scroll.

nullobject: Thanks for catching that, I will fix this in a follow-up PR.

PR #68, opened 2019-05-16
Fix placement issues with DonationDialog component

nullobject:
The modal is now horizontally & vertically centred on all screen sizes.

It will also grow vertically to fit the content, which means the user must scroll the page to see the dialog. This is better than the default Material UI behaviour, which sets a max height on the modal and creates little scroll bars inside it.

It would be good to make this the default behaviour for all of our dialogs in UI (@ZoeJazz and I were chatting about this yesterday), but let's get the behaviour right for the `DonationDialog` and then we can pull it up into our own `Dialog` component later.

## How to Test

- Check the `DonationDialog` at a variety of screen sizes and confirm the scrolling behaviour.

localhost_9001_iframe html_id=promos--donationdialog(iPhone 6_7_8)
localhost_9001_iframe html_id=promos--donationdialog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,10 @@
 ## Unreleased
 
+* Fix placement issues with `DonationDialog` component
+
 ## 2.2.3 (2019-05-14)
 
-* Update donation dialog breakpoints
+* Update `DonationDialog` breakpoints
 
 ## 2.2.2 (2019-05-09)
 
--- a/src/promo/DonationDialog.jsx
+++ b/src/promo/DonationDialog.jsx
@@ -1,28 +1,22 @@
-import MaterialDialog from '@material-ui/core/Dialog'
-import MaterialIconButton from '@material-ui/core/IconButton'
-import MaterialDialogActions from '@material-ui/core/DialogActions'
 import CloseIcon from '@material-ui/icons/Close'
+import MaterialIconButton from '@material-ui/core/IconButton'
 import PropTypes from 'prop-types'
 import React from 'react'
 import withStyles from '@material-ui/core/styles/withStyles'
+
+import Dialog from '../dialog/Dialog'
 import DialogActions from '../dialog/DialogActions'
 import Button from '../Button'
 
 const styles = theme => ({
+  root: {
+    display: 'flex',
+    overflowY: 'auto'
+  },
   container: {
-    // For small-medium devices, render the dialog container normally.
-    [theme.breakpoints.down('sm')]: {
-      display: 'block',
-      overflowY: 'auto',
-      overflowX: 'hidden'
-    },
-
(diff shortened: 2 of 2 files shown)


nullobject:
[[Bug] Fix placement issues with DonationDialog component](https://trello.com/c/mFGRQ8Xx/105-bug-fix-placement-issues-with-donationdialog-component)

nickbrowne:
Will check this as soon as I'm back from lunch :fork_and_knife:

nickbrowne:
Just gave this a go, it has some odd behaviour for me when scrolling in both Firefox and Chrome though.

Basically:
 - can't scroll "window" with mouse wheel in the blank part of the screen (previously could)
 - can't interact with scroll bar at all because it has an overlay on top of it

I made a few admittedly pro hacks to work around it, which does seem to work:
```diff
diff --git a/src/promo/DonationDialog.jsx b/src/promo/DonationDialog.jsx
index 5052f65..cf28e70 100644
--- a/src/promo/DonationDialog.jsx
+++ b/src/promo/DonationDialog.jsx
@@ -11,7 +11,12 @@ import Button from '../Button'
 const styles = theme => ({
   root: {
     display: 'flex',
-    overflowY: 'auto'
+    overflowY: 'auto',
+    backgroundColor: "rgba(0, 0, 0, 0.5)"
+  },
+  backdropRoot: {
+    backgroundColor: "unset",
+    position: "static",
   },
   container: {
     height: 'auto', // Override height
@@ -90,6 +95,11 @@ export const DonationDialog = ({
     open,
     onClose,
     onEnter: onVisible,
+    BackdropProps: {
+      classes: {
+        root: classes.backdropRoot
+      }
+    },
     classes: {
       root: classes.root,
       container: classes.container,
```

But while this fixes the scrolling issue, it breaks being able to close the dialog by clicking the `Backdrop`. Which made me start wondering if deviating too much from the systems that material has built up around how dialogs work is biting us, and might make upgrading more difficult in the future.

Is there a way we can achieve the layout we want without overriding as much of the default behaviour?

nullobject:
> Is there a way we can achieve the layout we want without overriding as much of the default behaviour?

@nickbrowne That's a shame. I agree, it feels like we are setting ourselves up for future pain by going against the MUI conventions. But I'm not sure how to achieve what we want using the default MUI `Dialog` 🤔

nickbrowne:
I might do a couple of little experiments after I finish up this branch I'm working on, hopefully it's possible!

nullobject:
> I might do a couple of little experiments after I finish up this branch I'm working on, hopefully it's possible!

That would be great, happy to chat/pair on it if you want.

jameshill:
Discussed with @nullobject, unless it comes up as a usability issue we're going to keep this on the backburner.

PR #71, opened 2019-06-04
[311] [Logo] Fix up logo image mistakes

markcipolla:
REFS https://trello.com/c/6HZVIFw7/311-logo-fix-up-logo-image-mistakes

@nickbrowne and I were looking at how we could pull these images into TC, and found a few small errors; the masthead images were the wrong size, and I'd goofed the footer images... mixed up which logo belongs to which language.

Still to investigate:

- [x] @nickbrowne had an inkling that AMP had strange requirements for logos and things. NEed to check that out, as I changed AMP to use the standardised set of images (same as the masthead)


nullobject:
Why did we end up prefixing the image filenames with `tcui`?

markcipolla:
> Why did we end up prefixing the image filenames with `tcui`?

Adding tcui to the filenames feels good to me, so when on TC I can know where that image is stored. Especially since we currently have some logos in TC-assets, and we have to differentiate between them while we feature flip them.

nullobject:
Fair enough. But from the perspective of the maintaining the UI package, it is introducing a convention which we don't use for other files in the package – to me that is surprising.

It seems like redundant information to include `tcui-` as a prefix for every file. Is there no other way we could solve this to play nicely with TC?

markcipolla:
> It seems like redundant information to include `tcui-` as a prefix for every file. Is there no other way we could solve this to play nicely with TC?

I'd originally made this to be a 1 for 1 drop-in replacement for TC-Assets. 

However, since we want to feature flip this, I had to rename the files. 
Plus, when including assets from outside your project, and doing so via the method we use in the ERB views (where webpack is tricked into `import`ing the images into the `manifest.json`), knowing where the images are is super helpful.

Not having the tcui-prefix is a pain when you're trying to figure out which logo is where.

nullobject:
Thanks for your response.

I'm going to leave the discussion in [here](https://github.com/conversation/tc/pull/9029) so that we're not commenting across two PRs 😜

markcipolla:
> I'm going to leave the discussion in [here](https://github.com/conversation/tc/pull/9029) so that we're not commenting across two PRs

giphy

PR #72, opened 2019-06-18
[394] [Logo] Regenerate large PNGs of logo for use on TC's republishing guidelines

markcipolla:
REFS https://trello.com/c/I3AKf4a3/394-logo-regenerate-large-png-images-for-republishing-guidelines

I had to regenerate the large PNG logo files, as they're not actually used to show the logo, but are available for download in the Republishing Guidelines.


RohanM:
I can confirm this PR increases the size of those images from `350 x 40` to `2030 x 232`. 

#### Before

before

#### After

after

PR #73, opened 2019-06-18
[96] Update UI for mobile newsletter sign up banner

markcipolla:
REFS https://trello.com/c/My5L0gL2/96-update-ui-for-mobile-newsletter-sign-up-banner

This PR duplicates the `DonationBanner`, which has a single use, and generalises it a bit, allowing for customisable colours ~of foreground and background text~ via theming!.

Screen Shot 2019-07-04 at 3 24 34 pm

To theme the banner, we'd wrap it in a `ThemeProvider` and associated theme, pick the `color` (`primary`, `secondary`, `default` or `inherit`), and off you go.

```
import React from 'react'
import { ThemeProvider } from '../../styles'
import ThinBanner from '../ThinBanner'
import coreTheme from '../../styles/themes/core'

<ThemeProvider theme={coreTheme()}>
  <ThinBanner color="primary" onClick={action('clicked')}>
    Stay informed with email updates
  </ThinBanner>

  <ThinBanner color="secondary" onClick={action('clicked')}>
    Stay informed with email updates
  </ThinBanner>
</ThemeProvider>
```


src/promo/ThinBanner.jsx
```diff
@@ -0,0 +1,106 @@
...
+export const ThinBanner = ({ children, backgroundColor, color, classes, onClick }) => {
+  let textColor
+
+  if (color) {
+    textColor = `${color}`
+  } else if (backgroundColor) {
+    // Break the hex colour into component colours
+    const hexcolour = hexRgb(backgroundColor)
+    // Counting the perceptive luminance human eye favors green color...
+    const brightness = 1 - (0.299 * hexcolour.red + 0.5 * hexcolour.green + 0.114 * hexcolour.blue) / 255
```
tiagoamaro: Would :heart: a comment with the source of this formula!

Found a StackOverflow comment about it https://stackoverflow.com/questions/596216/formula-to-determine-brightness-of-rgb-color#answer-596243 with a w3.org reference ("Techniques For Accessibility Evaluation And Repair Tools", https://www.w3.org/TR/AERT/#color-contrast).

Also, this could be a function :)

```js
/**
 * Perceived luminance for sufficient contrast
 * (https://www.w3.org/TR/AERT/#color-contrast)
 *
 * @param {String} color
 * @return {number}. Number between 0 and 1
 */
const perceivedLuminance = color => {
  const hexcolour = hexRgb(color)
  return 1 - (0.299 * hexcolour.red + 0.587 * hexcolour.green + 0.114 * hexcolour.blue) / 255
}
```

src/promo/ThinBanner.jsx
```diff
@@ -0,0 +1,106 @@
...
+
+    if (brightness < 0.5) {
+      // Show a dark text color
+      textColor = '#383838'
+    } else {
+      // Show a light text color
+      textColor = '#fff'
```
tiagoamaro: Should these `#ffffff` and `#383838` be part of the "theme"? If they're isolated cases, could live on constants, such as:

```js
const DARK_TEXT_COLOR = '#383838'
const LIGHT_TEXT_COLOR = '#ffffff'
```

src/promo/ThinBanner.jsx
```diff
@@ -0,0 +1,106 @@
...
+ *
+ * ~~~js
+ * <ThinBanner>
+ *   Some nice text
+ * </ThinBanner>
+ * ~~~
+ *
+ */
+export const ThinBanner = ({ children, backgroundColor, color, classes, onClick }) => {
+  let textColor
```
tiagoamaro: Mixing up my other comments, suggestion (if the text color is desired as a prop):

```js
const DARK_TEXT_COLOR = '#383838'
const LIGHT_TEXT_COLOR = '#ffffff'
const LUMINANCE_THRESHOLD = 0.5

const textColor = color ? color :
  perceivedLuminance(backgroundColor) < LUMINANCE_THRESHOLD ? DARK_TEXT_COLOR : LIGHT_TEXT_COLOR
```

src/promo/ThinBanner.jsx
```diff
@@ -0,0 +1,106 @@
...
+      <ArrowRightIcon className={classes.icon} />
+    </Button>
+  )
+}
+
+ThinBanner.propTypes = {
+  /**
+   * A background colour to override the default Red
+   */
+  backgroundColor: PropTypes.string,
```
tiagoamaro: Since the `backgroundColor` is not required, I see you're verifying if it's `undefined`. Could use `defaultProps` for this:

```js
import theme from "../styles/themes/default"

ThinBanner.defaultProps = {
  backgroundColor: theme().palette.core.main,
}
```

src/promo/ThinBanner.test.jsx
```diff
@@ -0,0 +1,71 @@
+import Button from '../Button'
+import ArrowRightIcon from '@material-ui/icons/KeyboardArrowRight'
+import ThinBanner from './ThinBanner'
+import React from 'react'
+import { mount, shallow } from 'enzyme'
```
tiagoamaro:
```suggestion
import { mount } from 'enzyme'
```

Since all `shallow` wrappers are being `.dive`d upon, you could simply `mount` them and replace all `wrapper.dive()` with `wrapper`.

ZoeJazz:
Heya, 

UI looks good. Nice work :)

Just a few things to note:
**- If you don't specify either, you get white text on TC red:** I wonder if we could apply the same colour logic as buttons for this? Defaults and variants should all match so we're not introducing new rules. Storybook isn't the best reference for buttons though as it's still in transition.
**- Should these #ffffff and #383838 be part of the "theme"?** Absolutely. We have labels for each of these colours that also hints at their usage (For instance, these are _neutral0_ and _neutral700_). The work to put these colours into storybook was halted due to material designs incompatible structuring of colours. I wonder if we need to pick this up again before wrapping up the banner work? Have attached the label structure. Plus, it's all in zeplin.
**- I wonder, if we're following W3 color contrast recommendation?** Based on [this check](https://docs.google.com/document/d/1_HK42qKBrlnY2F1YXACOyAV32YzhEhmqnJnu9KjzWsA/edit?usp=sharing) of all brand colours, I don't think #02A371 / positive600 meets AA requirements. I'd probably use core600, accent700, neutral700, neutral600 and positive700 if we're keen on having a whole set of options but we could just stick with the blue in the design (accent700) and iteratively develop things from there. Could we perhaps scrap the green and grey for this PR?

Colour labels

CC @tiagoamaro @markcipolla

src/promo/ThinBanner.jsx
nullobject: Magic numbers! ✨

nullobject:
@markcipolla Nice work. I agree that our UI library should contain components that are as reusable as possible. So if we can introduce something more generic that covers all our specific uses (e.g. `ThinBanner` vs. `ThinDonationBanner`), then I am all for it 👍 

@tiagoamaro Raises an interesting point about how we deprecate the other components. My guess would be that we rewrite them to use the new `ThinBanner` internally, and mark them as deprecated.

@ZoeJazz Made some interesting points about our default colours. For buttons, I think it makes sense that the default colour isn't the primary or secondary color. This is a convention carried over from the MUI library.

Maybe we should just do the same with the `ThinBanner`? The default colour could be the neutral grey background, just like our buttons. Everything else is more of a modifier to the default state.

The other point that @ZoeJazz raised was to do with our colours. Ideally we should just have a predefined palette of colours that we refer to semantically in our code. For example, `core`, `aware`, `positive`, rather than `#02a371`. I don't think that work precludes this PR, but it's something we need to solve.

ZoeJazz:
> if we can introduce something more generic that covers all our specific uses (e.g. ThinBanner vs. ThinDonationBanner), then I am all for it 👍

> The default colour could be the neutral grey background, just like our buttons. Everything else is more of a modifier to the default state.

100% agree @nullobject

src/promo/stories/ThinBanner.jsx
```diff
@@ -0,0 +1,28 @@
...
+
+    <ThinBanner backgroundColor='#48509e' onClick={action('clicked')}>
+      Stay informed with email updates
+    </ThinBanner>
+
+    <ThinBanner backgroundColor='#43a047' onClick={action('clicked')}>
+      Let us know what you think by taking part in our yearly survey
+    </ThinBanner>
+
+    <ThinBanner backgroundColor='#eeeeee' Click={action('clicked')}>
```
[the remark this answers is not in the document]
markcipolla: Fixed 🙏

src/promo/ThinBanner.jsx
markcipolla: Yes. Extracted to a function. Also left a comment with the URL to the stackoverflow where those magic numbers came from, so future-us can figure out what our intent was.

nullobject:
@markcipolla Hey mate, Zoe told me that you're going to have a crack at adding semantic colours. Sounds like a good move, let me know if you want a hand with that, either in this PR or a new one 👍

markcipolla:
> @markcipolla Hey mate, Zoe told me that you're going to have a crack at adding semantic colours. Sounds like a good move, let me know if you want a hand with that, either in this PR or a new one 👍

Yes! @ZoeJazz and I were chatting, and then @nickbrowne and I had a chat. I'll defs be picking your brain on this; I'll schedule a meeting for Monday / Tuesday next week where we can have a discussion on it.

@nickbrowne has been showing m how the theming works, and I think we might have a nicer solution, but I'll spike it out and see.

markcipolla:
Now that https://github.com/conversation/ui/pull/75 merged, using the themes is now super easy! I've reworked this PR, removing all the custom colour generation stuff, and instead relying on wrapping the component in a `ThemeProvider`.

Please review, have a look, and tell me what y'all think.

markcipolla:
> @markcipolla , noticed that the `ThinBanner` component is not being publicly available on the module namespace (not exported on `src/index.js`). Is this intentional?

Ah, no it was not intentional. I'll add.

> Not for this PR, but noticed that now the default theme uses MUI's "neutral" colors, some promo components that used the previous "core" theme are mistyled:

I'll look at updating the components with themes, and we can iterate on them to make sure they're as we want.

markcipolla:
> @markcipolla , noticed that the `ThinBanner` component is not being publicly available on the module namespace (not exported on `src/index.js`). Is this intentional?

Added

PR #75, opened 2019-06-25
[432] [SPIKE] Update TCUI default theme, and add requried additional themes/colours

markcipolla:
## This PR is for discussion first, before we merge anything. We're unsure of the best way to theme components, and this is only one method.

This PR basically tries to turn this:
59724216-75ae6f00-926c-11e9-9044-972e48e9d23d

into this:
Screen Shot 2019-06-27 at 5 14 42 pm

So far, it's only the default `neutral` theme, and adds additional `core` and `accent` colours to new themes.

Material UI wants us to provide colours to things via a 'theme'. The theme has a primary and secondary colour, and when applied to a component, uses those colours to set the component colour + text colour (which is made sure to contrast enough).

Eventually, to get the TC brand red (which we have as the `primary` colour in our 'core' theme) on a `Banner` example component, it'll be as easy as:

```
<ThemeProvider theme={coreTheme()}>
  <Banner color="primary">Some banner text!</Banner>
</ThemeProvider/
```

We still need to wrangle the colours, but I think it'll be a good discussion point ahead of building theme-able UI components.


markcipolla:
Screen Shot 2019-06-25 at 4 20 14 pm
I've added an example showing how the buttons would be themed.

ZoeJazz:
Thanks @markcipolla 

We probably wouldn't have a need for all these button variations. It's not a very colourful brand. 

The colour model works like this:
- Neutral: Default for everything ie font colour, UI component borders, backgrounds, buttons, etc
- Core: Brand reinforcement ie promo messaging, logo icon, etc
- Accent: Highlighted areas of the UI ie promo messaging, calls to action, etc
- Positive: UI Feedback for successful form validation or 'on' state for radio buttons and switches
- Aware: Background colour for alerts (if necessary / very low use case)

z

Screen Shot 2019-06-25 at 4 58 18 pm
Screen Shot 2019-06-25 at 4 58 24 pm
Screen Shot 2019-06-25 at 5 03 06 pm

src/styles/stories/colours.jsx
```diff
@@ -1,19 +1,95 @@
...
+import { ThemeProvider } from '../'
+import { Swatch, SwatchGroup } from '../../Swatch'
+import Typography from '../../Typography'
+
+import accentTheme from '../themes/accent'
+import awareTheme from '../themes/aware'
+import coreTheme from '../themes/core'
+import defaultTheme from '../themes/default'
+import neutralTheme from '../themes/neutral'
+import positiveTheme from '../themes/positive'
```
tiagoamaro: Got the idea of breaking themes apart! Great way on working MUI's limitation on multiple colors :clap: 

I wonder how we could use this with other components, when importing `@theconversation/ui` while still wrapping them within the `ThemeProvider`. Example, if I want to create a "neutral" donation banner, I would like to use the following API:

```jsx
<ThemeProvider theme="neutral">
  <DonationBanner>
    The Conversation provides clean information in a world infected with spin. Please donate.
  </DonationBanner>
</ThemeProvider>
```

Right now, themes are not exported at the `src/index.js`, and IMHO, they shouldn't! (cc @nullobject @nickbrowne), but if they're are available for wider usage, they should be part of a configurable prop.

----

If this API gets through, and `theme` is a configurable prop with a subset, the `ThemeProvider` component should be able to handle it:

```js
// At the `src/styles/ThemeProvider.jsx` file propTypes definitions
theme: PropTypes.oneOf(['default', 'core', 'accent']), // and others....
```

nullobject:
Very interesting, I didn't think of trying to solve it this way 🤔 

I was thinking something more along the lines of this: https://stackblitz.com/edit/react-bd92xq

Here they are just adding new colours to the theme, in addition to `primary` and `secondary`.

nullobject:
This was the discussion on Stack Overflow: https://stackoverflow.com/questions/50069724/how-to-add-multiple-material-ui-palette-colors

nullobject:
...but I really the promise of this PR, it's a much better way of solving the colour issues.

I would like to understand how these different themes would map to our forms. Donations is probably a good example, because it has been moved over to use TC UI.

src/styles/stories/colours.jsx
markcipolla: This makes each component responsible for wrapping _itself_ in a theme component?

The thinking with this was that we would leave the wrapping of a component in a theme to where it's used, alleviating the components themselves of having to know anything about what colours to use.

src/styles/stories/colours.jsx
```diff
@@ -1,19 +1,95 @@
...
+  return (
+    <React.Fragment>
+      <Typography style={{ clear: 'both' }} variant='h4'>Primary</Typography>
+      <ThemeProvider theme={defaultTheme()}>
+        <SwatchGroup groupLabel={'Default'}>
+          {COLOURS.map((colour) =>
+            <div>
+              {VARIANTS.map((variant, index) =>
+                <Swatch colour={colour} variant={variant} />
+              )}
```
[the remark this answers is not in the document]
markcipolla: Good call.

src/Swatch/index.js
```diff
@@ -1 +1,2 @@
-export { default } from './Swatch'
+export { default as Swatch } from './Swatch'
+export { default as SwatchGroup } from './SwatchGroup'
```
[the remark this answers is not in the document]
markcipolla: I'd say that neither should be included in `src/index.js` as they'll not be used outside of the storybook.

markcipolla:
## Changes to our current theme:

### From:
Screen Shot 2019-06-27 at 5 16 23 pm

### To:
Screen Shot 2019-06-27 at 5 16 16 pm

src/styles/stories/colours.jsx
tiagoamaro: Yeah, not the right call! @nullobject proposal is way better!

src/Swatch/index.js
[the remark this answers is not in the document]
tiagoamaro: @markcipolla , FYI, opened a PR after thinking about this: https://github.com/conversation/ui/pull/76.

ZoeJazz:
Yep :) That works well @markcipolla

markcipolla:
Does anything else need to be addressed in this PR? 
Are we happy to merge it and give it a try?

ZoeJazz:
From a purely aesthetic perspective, this solution looks great. If everyone is happy with how the resulting codebase is simplified and made more adaptive, then I'm in full support of this solution. Well done @markcipolla. Really great work.

src/Swatch/Swatch.jsx
```diff
@@ -8,13 +8,15 @@ const size = 120
 const margin = 16
 const styles = {
   root: {
+    display: 'block',
+    bacground: '#eee',
```
benmacleod:
```suggestion
    background: '#eee',
```

src/styles/themes/common.js
```diff
@@ -0,0 +1,36 @@
...
+
+  overline: {
+    letterSpacing: 0.25,
+    textTransform: 'none'
+  }
+}
+
+export const errorStateColours = {
+  palette: {
+    error: { main: '#990000', border: '#d8352a' }
```
benmacleod: At the risk of prolonging a super long-running PR, and with apologies if this has been suggested, but how would it be to extract the hard-coded colour names to a file of constants? It would be great if they had semantic names rather than just being "magic".

benmacleod: They could be an export from this very file in fact.

src/Swatch/Swatch.jsx
markcipolla: Fixed.

src/styles/themes/common.js
markcipolla: > It would be great if they had semantic names

What would be good names?

benmacleod: > What would be good names?

If I read it right, e.g. `#990000` is colour "Core", strength 700. So, maybe `colours.core700` or something?

markcipolla: HMMMM yeah, that feels pretty good... I'll follow the https://material-ui.com/customization/color/#color style... so core: {"700": "#990000"}?

Something like the example below?

```
import purple from '@material-ui/core/colors/purple';
import red from '@material-ui/core/colors/red';

const primary = red[500]; // #F44336
const accent = purple['A200']; // #E040FB
const accent = purple.A200; // #E040FB (alternative method)
```

benmacleod: Yeah, that sounds good. As long as you use the same names as the style guide, so, e.g. 'core', 'accent', etc.

markcipolla: Done 🎉

src/styles/palettes/accent.js
```diff
@@ -0,0 +1,7 @@
+const accent = {
```
benmacleod: 👏

tiagoamaro:
I'm :100:% with this approach now! :tada:

Here are some screenshots with the three themes applied to the date picker:

<details>

- Accent theme

image

- Core theme

image

- Default theme

image

</details>

tiagoamaro:
Let's get this shipping! Incredible work, discussion and results @markcipolla :trophy: :tada:

ZoeJazz:
Performing exactly as intended! Thanks @markcipolla for persevering with the new colour palette.

benmacleod:
Happy to see this go in - 👏 @markcipolla

PR #76, opened 2019-06-27
Do not export the Swatch component

tiagoamaro:
This PR removes the `Swatch` component from the package namespace, as it's a component to showcase theme styling. It shouldn't be used outside Storybook.


benmacleod:
Yep, sounds sensible 👍

PR #78, opened 2019-07-05
[517] Fix existing TCUI donation related banners

markcipolla:
### Relies on https://github.com/conversation/ui/pull/81 being merged first

When we added the new themes, we naffed the donation-related banners' styling a smidge. This was because we changed the default colours and how we structured the themes.

I've made them look as expected again. Most of these are pulling in the `core` palette, and specifying the colour manually. The ones where we were setting backgrorund colours required this method.

One I could get away with using a `ThemeProvider` was in the below-article banner, as only the button needed colouring.

I think this is okay for now; we built those components pre-themes, so keeping them the same for now while we can generalise them for future use (like in the case of making the `ThinBanner`, which was a generalisation of the `ThinDonationBanner`).

# Left is this PR, right is current master

Screen Shot 2019-07-05 at 3 11 52 pm (2)
Screen Shot 2019-07-05 at 3 11 57 pm (2)
Screen Shot 2019-07-05 at 3 12 02 pm (2)
Screen Shot 2019-07-05 at 3 12 05 pm (2)
Screen Shot 2019-07-05 at 3 12 09 pm (2)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased
+
+* Update donation components to use themes
+
 ## 2.6.0 (2019-07-09)
 
 * Update Material UI to 4.2.0 (https://github.com/conversation/ui/pull/81)
--- a/src/promo/DonationBanner.jsx
+++ b/src/promo/DonationBanner.jsx
@@ -1,7 +1,7 @@
 import Button from '../Button'
 import CloseIcon from '@material-ui/icons/Close'
 import MaterialIconButton from '@material-ui/core/IconButton'
-import Paper from '@material-ui/core/Paper'
+import Box from '@material-ui/core/Box'
 import PropTypes from 'prop-types'
 import React from 'react'
 import Typography from '../Typography'
@@ -10,9 +10,7 @@ import neutralPalette from '../styles/palettes/neutral'
 import withStyles from '@material-ui/core/styles/withStyles'
 
 const styles = theme => ({
-  // TODO: this is temporary while we work out what to do about
-  // themes and palettes properly
-  paper: {
+  box: {
     position: 'relative',
     backgroundColor: corePalette[600],
     color: neutralPalette[0],
@@ -35,9 +33,6 @@ const styles = theme => ({
   // Donation button styles are a special case until we decide
   // "inverted" buttons are going to be a more widespread thing
   button: {
-    // We need to mark the colour with `!important` to avoid global hyperlink
-    // styles overriding this value.
-    color: [theme.palette.text.primary, '!important'],
     backgroundColor: neutralPalette[0],
     fontWeight: 'bold',
     '&:hover': {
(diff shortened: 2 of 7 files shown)


src/promo/DonationBanner.jsx
```diff
@@ -5,15 +5,16 @@ import Paper from '@material-ui/core/Paper'
...
 
 const styles = theme => ({
   // TODO: this is temporary while we work out what to do about
   // themes and palettes properly
   paper: {
     position: 'relative',
-    backgroundColor: theme.palette.core && theme.palette.core.main,
-    color: theme.palette.primary.contrastText,
+    backgroundColor: corePalette[600],
+    color: '#fff',
```
nickbrowne: In hindsight we might be better off using the more basic `Box` instead of `Paper` as the main wrapper of this component.

I think it would let us do away with most of these style overrides:
```jsx
<Box bgcolor="primary.main" color="primary.contrastText" p={4}>
  Some cool content
</Box>
```
That way we're not fighting the border radius/elevation either.

ZoeJazz: Could this be color: neutralPalette[0] ?

src/promo/DonationDialog.jsx
```diff
@@ -27,9 +28,9 @@ const styles = theme => ({
   paper: {
     // TODO: revisit palette colours with what we learned building
     // this component.
-    backgroundColor: theme.palette.core && theme.palette.core.main,
+    backgroundColor: corePalette[600],
     overflow: 'visible', // So we can have the avatar poking out the top
-    color: theme.palette.primary.contrastText,
+    color: '#fff',
```
ZoeJazz: And here too. 

color: neutralPalette[0]

src/promo/DonationBanner.jsx
markcipolla: Yes. It defs can. 👏

markcipolla: Good call; switching to `Box` over `Paper` means we don't need to override `elevation` nor `borderRadius`

src/promo/DonationDialog.jsx
markcipolla: Aaaand done.

markcipolla:
Ah, bumping `@materual-ui/core` has broken much.

markcipolla:
> Ah, bumping `@materual-ui/core` has broken much.

Hooray for https://github.com/conversation/ui/pull/81 unblocking this PR!

src/promo/ThinDonationBanner.jsx
```diff
@@ -3,20 +3,21 @@ import ArrowRightIcon from '@material-ui/icons/KeyboardArrowRight'
...
   button: {
     borderRadius: 0,
     justifyContent: 'left',
-    backgroundColor: theme.palette.core && theme.palette.core.main,
+    backgroundColor: corePalette[600],
 
     // We need to mark the colour with `!important` to avoid global hyperlink
     // styles overriding this value.
-    color: [theme.palette.primary.contrastText, '!important'],
+    color: ['#fff', '!important'],
```
RohanM: Could/should this use `neutralPalette[0]` here?

markcipolla: Ahhh, must've messed up a rebase. Yes.

markcipolla: @RohanM Done. Force pushed

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## Unreleased
+
+* Update donation components to use themes
```
markcipolla: 👏

PR #82, opened 2019-07-12
[514] Build a component (MessageTile) to show the republish panel in TCUI

markcipolla:
REFS https://trello.com/c/gBzlSziL/514-build-republish-component-in-tcui

This PR adds a `MessageTile` component, which takes optional copy-related props to show promo messages.

It's quite configurable it it's use; the one component could be purposed for a number of promos. The storybook shows a few options.

Screen Shot 2019-07-12 at 2 25 58 pm

I've also added small style changes to the typography, closer aligning the font sizes / letter spacings from @ZoeJazz's designs.


src/styles/themes/common.js
```diff
@@ -8,14 +8,25 @@ export const typography = {
   h3: { fontFamily: 'Montserrat', fontWeight: 700 },
   h4: { fontFamily: 'Montserrat', fontWeight: 600 },
   h5: { fontFamily: 'Montserrat', fontWeight: 600 },
-  h6: { fontFamily: 'Montserrat', fontWeight: 700 },
+  h6: {
+    fontFamily: 'Montserrat',
+    fontWeight: 700,
+    lineHeight: 'normal',
+    letterSpacing: '0.25px'
+  },
```
tiagoamaro: > I've also added small style changes to the typography, closer aligning the font sizes / letter spacings from @ZoeJazz's designs. - @markcipolla 

:+1:

src/promo/stories/messageTile.jsx
```diff
@@ -0,0 +1,72 @@
...
+              <img style={{ width: 26, marginRight: 6, marginBottom: 8 }} src={by} />
+              <img style={{ width: 26, marginRight: 6, marginBottom: 8 }} src={cc} />
+              <img style={{ width: 26, marginBottom: 8 }} src={nd} />
+              <div>We believe in the free flow of information.</div>
+            </React.Fragment>
+          }
+          body='All our articles can be republished for free, online or in print, under the Creative Commons licence.'
+          buttonText='Republish for free'
+          onClick={action('clicked')}
+        />
```
tiagoamaro: The `MessageTile` usage reminds me of MUI's `Dialog`. There are components within it which have to have their specific styles and events, but they should be flexible enough to support components within them. I don't think that `MessageTile`'s inner components (`title`, `body`, `button`) should be passed as props.

What do you think about this API?

```jsx
<MessageTile>
  <MessageTileHeader>
    <img style={{ width: 26, marginRight: 6, marginBottom: 8 }} src={by} />
    <img style={{ width: 26, marginRight: 6, marginBottom: 8 }} src={cc} />
    <img style={{ width: 26, marginBottom: 8 }} src={nd} />
    <div>We believe in the free flow of information.</div>
  </MessageTileHeader>
  <MessageTileBody>
    All our articles can be republished for free, online or in print, under the Creative Commons licence.
  </MessageTileBody>
  <MessageTileButton onClick={action('clicked')}>
    Republish for free
  </MessageTileButton>
</MessageTile>
```

With this, you would have the root component (`MessageTile`) with its three other sub-components: `MessageTile/MessageTileHeader`, `MessageTile/MessageTileBody` and `MessageTile/MessageTileButton`.

This approach also avoids mixing button related props with the root component. That could cause a bit of confusion, as all inner components are optional.

What do you think?

markcipolla: This sounds pretty good. I'd thinking I'd like to keep it in the `/src/promo` folder... but yeah, now I see the similarities to the `Dialog`.

src/promo/MessageTile.jsx
```diff
@@ -0,0 +1,74 @@
...
+
+import Button from '../Button'
+import Typography from '../Typography'
+import neutralPalette from '../styles/palettes/neutral'
+
+const styles = theme => ({
+  body: {
+    marginBottom: 16
+  },
+  box: {
```
ZoeJazz: Are these repeatable attributes? All measures will be divisible by 4 so it's great if we can consolidate the spacing somehow.

nickbrowne: If our baseline spacing unit is 4 then we should really configure that in the theme. Currently our baseline spacing unit is the default (8 pixels).

Changing to 4 may mess with a lot of the default components and third party libraries though, so we'll want to be careful when changing it. May not be an issue though - I haven't tried it.

Either way we should aim to use the theme spacing methods wherever possible instead of magic numbers.

markcipolla: I don't think we should change the spacing unit from 8px to 4px; MUI specifically recommends using a 8px spacing https://material-ui.com/customization/spacing/

I agree with Nick that it'll probably quite negatively affect other components.

But yes, I'll change it to not have magic numbers around. We can always do `theme.spacing(0.5)` for 4px.

src/promo/stories/messageTile.jsx
markcipolla: I've revised my opinion, and I agree putting it in`src/MessageTile` does make sense...
We have some components in their own top level folder, and some in subfolders.

src/promo/MessageTile.jsx
ZoeJazz: Ah yes - Let's keep to 8.

PR #83, opened 2019-07-15
[559] Add Box, themes and palettes to exports from TCUI

markcipolla:
REFS https://trello.com/c/9mlHdoKu/559-add-box-our-themes-and-palettes-to-tcui

I noticed that we could export a few more things, especially the themes and palettes; the themes we will need to wrap around other TCUI components to adjust their styling, and the palettes for when we need to use standardised colours..

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Unreleased
 
+* Export MUI `Box`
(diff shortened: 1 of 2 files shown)


markcipolla:
> Also, TIL that MUI's `Box`es are basically powerful `div`s.

Yep, Divs with extra props!

PR #88, opened 2019-08-13
Lock material-ui/core to 4.2.0

markcipolla:
https://trello.com/c/SZC5hsy6/783-lock-tcui-to-material-ui-core-420

This PR locks `@material-ui/core` to 4.2.0 and downshift to 3.2.7, and sets specific versions to material-ui related packages.

@nickbrowne and I looked at a js package upgrade on TC. on TC, Most of the changes came from TCUI. This PR locks those versions down (removes the `^`).

Then, upgrades should be by a choice, and then pushed to TC.
When we bump material-ui/core to 4.3.2 in a future PR, downshift needs to upgrade to v4.0.0 then.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1579,25 +1579,25 @@
       }
     },
     "@material-ui/icons": {
-      "version": "4.2.1",
-      "resolved": "https://registry.npmjs.org/@material-ui/icons/-/icons-4.2.1.tgz",
-      "integrity": "sha512-FvSD5lUBJ66frI4l4AYAPy2CH14Zs2Dgm0o3oOMr33BdQtOAjCgbdOcvPBeaD1w6OQl31uNW3CKOE8xfPNxvUQ==",
+      "version": "4.2.0",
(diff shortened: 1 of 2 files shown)


nickbrowne:
@nullobject have you got any objections to doing this?

What we found was that for apps including `@theconversation/ui`, they would often have minor versions quite different from what we were actually developing `@theconversation/ui` with.

eg. While currently we've got 4.2.0 in `package-lock.json` here, the loose dependency in the `package.json` means that installing it from TC today will end up with 4.3.2 (the latest), which does change behaviour slightly.

This is not to say we shouldn't upgrade material ui, more that it becomes an intentional thing where we bump the version here and that flows through.

markcipolla:
> This is not to say we shouldn't upgrade material ui, more that it becomes an intentional thing where we bump the version here and that flows through.

☝️ very much this. Thank you for putting it more eloquently

markcipolla:
@nickbrowne @nullobject It follows on to https://github.com/conversation/ui/pull/89

Does this sound a reasonable plan?

nickbrowne:
> Yep, this is a good idea.

Thanks

PR #91, opened 2019-08-26
Allow Person to optionally render a caption if prop provided

markcipolla:
https://trello.com/c/pqLwhpmI/873-make-person-accept-an-optional-caption-prop

The TCF donation campaign has thrown a lil' curveball; they wanted a sign off on an `EndOfArticle` from the whole team!

Screen Shot 2019-08-26 at 10 39 30 am

That's cool, but it causes a few errors as the `Person` component expects a `caption` prop, and Material-UI doesn't like to render empty `Typography` components.

While there, I added `propTypes` and `defaultProps`, made the docs more in-line with our style (using `ComponentOverview` and putting the docs on the component itself).


markcipolla:
Screen Shot 2019-08-26 at 4 51 04 pm

It also improves the storybook page.

PR #97, opened 2019-09-13
Show more examples of MessageTile components + enhance MessageTile

markcipolla:
https://trello.com/c/QhX47tvY/991-show-more-examples-of-messagetile-components

A point raised in a team meeting was that we should show better examples of how our newer configurable components can be used.

We're aiming to deprecate the older, hardcoded promos:
- `ArticleDonationBanner`
- `DonationBanner`
- `DonationDialog`
- `ThinDonationBanner`

In favour of using newer, configurable components:
- `ThinBanner`
- `ThickBanner`
- `MessageTile`

This PR will tidy up the examples of MessageTile, and **enhance it to fully-replicate** the `EndOfArticle` component (which used `ArticleDonationBanner` under the hood, but now can be done with `MessageTile`). 

The enhancements add two props to `MessageTileButton` to enable it to meet all the requirements we currently have for it.

Screen Shot 2019-09-13 at 12 07 08 pm

### DIFF
--- a/src/MessageTile/MessageTileButton.jsx
+++ b/src/MessageTile/MessageTileButton.jsx
@@ -9,9 +9,9 @@ const styles = theme => ({
   }
 })
 
-export const MessageTileButton = ({ children, fullWidth, prominent, classes, onClick }) => {
+export const MessageTileButton = ({ children, classes, ...props }) => {
   return (
-    <Button fullWidth={fullWidth} prominent={prominent} className={classes.button} colour='primary' onClick={onClick}>
+    <Button className={classes.button} colour='primary' {...props}>
       {children}
     </Button>
   )
@@ -24,17 +24,9 @@ MessageTileButton.propTypes = {
    */
   children: PropTypes.node.isRequired,
   /**
-   * The onClick event for the Button
+   *
    */
-  onClick: PropTypes.func,
-  /**
-   * If the button should run full width
-   */
-  fullWidth: PropTypes.bool,
-  /**
-   * If the button should be a little wider, having extra padding left and right
-   */
-  prominent: PropTypes.bool
+  classes: PropTypes.object
 }
 
 MessageTileButton.defaultProps = {
--- a/src/MessageTile/MessageTileButton.jsx
+++ b/src/MessageTile/MessageTileButton.jsx
@@ -9,9 +9,9 @@ const styles = theme => ({
   }
 })
 
-export const MessageTileButton = ({ children, classes, onClick }) => {
+export const MessageTileButton = ({ children, fullWidth, prominent, classes, onClick }) => {
   return (
-    <Button fullWidth className={classes.button} colour='primary' onClick={onClick}>
+    <Button fullWidth={fullWidth} prominent={prominent} className={classes.button} colour='primary' onClick={onClick}>
       {children}
     </Button>
   )
@@ -26,7 +26,20 @@ MessageTileButton.propTypes = {
   /**
    * The onClick event for the Button
    */
-  onClick: PropTypes.func
+  onClick: PropTypes.func,
+  /**
+   * If the button should run full width
+   */
+  fullWidth: PropTypes.bool,
+  /**
+   * If the button should be a little wider, having extra padding left and right
+   */
+  prominent: PropTypes.bool
+}
+
+MessageTileButton.defaultProps = {
+  fullWidth: true,
+  prominent: false
 }
 
 export default withStyles(styles)(MessageTileButton)
--- a/src/MessageTile/stories/examples.jsx
+++ b/src/MessageTile/stories/examples.jsx
@@ -1,0 +1,137 @@
+import React from 'react'
+import { ThemeProvider } from '../../styles'
+import Typography from '../../Typography'
+import MessageTile from '../MessageTile'
+import MessageTileHeader from '../MessageTileHeader'
+import MessageTileBody from '../MessageTileBody'
+import MessageTileButton from '../MessageTileButton'
+import { action } from '@storybook/addon-actions'
+
+import Avatar from '../../Avatar'
+import Person from '../../Person'
+
+import by from './by.svg'
+import cc from './cc.svg'
+import nd from './nd.svg'
+
+import accentTheme from '../../styles/themes/accent'
+import coreTheme from '../../styles/themes/core'
+import defaultTheme from '../../styles/themes/default'
+
+export default () => (
+  <div>
+    <Typography
+      style={{ clear: 'both', margin: '20px 0' }}
+      variant='h4'
+    >
+      Republishing
+    </Typography>
+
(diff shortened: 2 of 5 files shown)


src/MessageTile/stories/messageTile.jsx
```diff
@@ -37,54 +34,6 @@ export default () => (
...
-            </MessageTileHeader>
-            <MessageTileBody>
-              Subscribe to our daily newsletter and get the latest analysis and commentary directly in your inbox.
-            </MessageTileBody>
-            <MessageTileButton onClick={action('clicked')}>
-              Subscribe now
-            </MessageTileButton>
-          </MessageTile>
-        </ThemeProvider>
-      </div>
```
tiagoamaro: Good move removing this from the "overview" :trophy:

src/MessageTile/MessageTileButton.jsx
```diff
@@ -26,7 +26,20 @@ MessageTileButton.propTypes = {
...
-  onClick: PropTypes.func
+  onClick: PropTypes.func,
+  /**
+   * If the button should run full width
+   */
+  fullWidth: PropTypes.bool,
+  /**
+   * If the button should be a little wider, having extra padding left and right
+   */
+  prominent: PropTypes.bool
```
tiagoamaro: `prominent` really makes it just a tiny bit wider. Had a hard time finding the difference when toggling this prop.

- With `prominent` `false`

image

- With `prominent` `true`

image

markcipolla:
It's subtle, but to get parity

src/MessageTile/MessageTileButton.jsx
ZoeJazz: Good point @tiagoamaro! It's unclear from a UI perspective why this is a prop.

src/MessageTile/stories/examples.jsx
```diff
@@ -0,0 +1,137 @@
...
+              willing to drill down to the truth. But we can't do that vital
+              work unless readers donate. Please make a donation.
+            </Typography>
+          </MessageTileBody>
+
+          <MessageTileButton fullWidth={false} prominent onClick={action('clicked')}>
+            Donate now
+          </MessageTileButton>
+
+          <div key='attribution'>
```
ZoeJazz: Code looks great! Does this follow the same pattern as the Dialog attribution?

src/MessageTile/stories/messageTile.jsx
```diff
@@ -12,13 +11,11 @@ import by from './by.svg'
...
 import nd from './nd.svg'
 
-import accentTheme from '../../styles/themes/accent'
-import coreTheme from '../../styles/themes/core'
 import defaultTheme from '../../styles/themes/default'
 
 export default () => (
   <ComponentOverview heading='MessageTile' component={UnwrappedMessageTile}>
-    <div style={{ width: 220 }}>
+    <div style={{ marginTop: 20, width: 220 }}>
```
ZoeJazz: Where is this margin sourced from? This might be an error in Zeplin as we are aiming to stick with divisions of 8.

markcipolla: These are examples only; to show how they could be used. The margins are only to separate the examples.

src/MessageTile/stories/examples.jsx
markcipolla: I think you mean the [DonationDialog](http://styleguide.theconversation.com/?path=/story/promos--donationdialog) component rather than the [Dialog
](http://styleguide.theconversation.com/?path=/story/dialogs--content), but they both use the `Person` component.

In this case, the MessageTile is separate from the attribution; as far as the MessageTile is concerned, it's just content. It could be anything.

src/MessageTile/MessageTileButton.jsx
markcipolla: It's a prop because the underlying component (the `Button`) accepts `prominent` as a prop. This makes the MessageTile be able to 100% replicate the design of the [ArticleDonationBanner](http://styleguide.theconversation.com/?path=/story/promos--articledonationbanner). My goal is to deprecate the ArticleDonationBanner, and instead use MessageTile.

If you're amenable to it, we could look at removing the distinction in `Button` (which is our code), and then not having it as a concept at all; the button on the end of article call-to-action would change ever so slightly (the button would be slightly less wide), but everything would be more uniform and consistent.

nickbrowne: If button accepts that prop already, can't a regular Button just be a child within MessageTile?

markcipolla: The `MessageTileButton` builds the button rather than accepts it as a child, attaching the `onClick` and adding styles (adding them via the theming).

A regular `Button` _can definitely_ be a child within `MessageTile`, but it won't pick up the margin.

markcipolla: If you have a better way to structure it, I'm all ears. 

Also, my opinion is to just remove `prominent` from `Button`, and reduce the number of options. Then, fewer differing types of buttons, and no need to include it as a configurable prop here.

I think it was added in 86caab8e38fbf88db766b86541358f35a1a59cf1.

tiagoamaro: > If button accepts that prop already, can't a regular Button just be a child within MessageTile? - @nickbrowne 

> The `MessageTileButton` builds the button rather than accepts it as a child... - @markcipolla 

### Suggestion

Since the `MessageTileButton` is a special MUI `Button` component, with custom `marginBottom` spacing, it could simply accept all props of a regular `Button`, so it would as flexible as possible whenever rendering them:

```jsx
export const MessageTileButton = ({ children, classes, ...props }) => {
  return (
    <Button {...props} className={classes.button} colour='primary'>
      {children}
    </Button>
  )
}

MessageTileButton.defaultProps = {
  fullWidth: true,
  prominent: false
}
```

Same behaviour, also having default props documented, but allowing all regular MUI `Button` props to be passed to it. Just would need to document that `className` and `colour` are constant values.

### Screenshot

image

markcipolla: @nickbrowne thoughts? I think that is the best of both worlds. 👏 Good idea @tiagoamaro !

> ...but allowing all regular MUI Button props to be passed to it. Just would need to document that className and colour are constant values.

Very nice. `MessageTileButton` goes back to knowing less about the implementation of `Button`, as it should.

Also, after chatting with @ZoeJazz and @jameshill, we have made a item in the [TCUI transition checklist](https://trello.com/c/xkAe8637/747-tcui-transition) to look at removing `prominent` from `Button`, and the components which also use it.

src/MessageTile/stories/messageTile.jsx
markcipolla: I've adjusted this to 16px; may as well follow a 8px base unit if we're going to use it for everything else.

PR #98, opened 2019-09-16
Omit the term ‘Promos’ from UI library by moving them to top-level components

markcipolla:
https://trello.com/c/kA39ccFr/997-omit-the-term-promos-from-ui-library

This PR is moves the `promos` folder of components into individual top-level components. We wanted to tidy up the styleguide. The sidebar is now like this:

Screen Shot 2019-09-16 at 12 05 26 pm

These four components are on the 🤞 soon-to-be-deprecated list, being replaced with configurablealternatives, so hopefully won't be around long.


ZoeJazz:
Great to see this work happening. Thanks @markcipolla

markcipolla:
🙇

PR #112, opened 2019-11-04
Change MessageTile's spacing to get parity with Zoe's designs

markcipolla:
- Change `MessageTile`'s padding to top/bottom 24px and left/right 16px
- Remove marginBottom from `MessageTileButton`
- I've also built a tiny Add a `MessageTileAttribution` component. We have a non-componentised repeated use of this attribution block, especially in TC; to ensure it's exactly as per designs, I've extracted it to a tiny component whose sole job is to give spacing between the `MessageTileButton` and the attribution, but also to lock down the styles; that it's exactly 16px / `theme.spacing(2)`.

This way, the entire `MessageTile` component (and all children components) exactly match the designs. 

Left is master, right is this PR. Note the padding is 16px 24px, the button doesn't push down the attribution, but instead the new attribution component gives _itself_ spacing. 
Screen Shot 2019-11-04 at 11 58 09 am (3)

This (apart from the heading change in https://github.com/conversation/ui/pull/111, gives us exact parity in spacing to @ZoeJazz's designs in https://gallery.io/projects/MCHbtQVoQ2HCZWiQBR82z4LY/files/MCEJu8Y2hyDScfjXeeKP3gqEElw_0mfRSGY


markcipolla:
If y'all feel this needs to be split up, it very much can, but the changs are pretty small and focused only on the one component; they're all needed to get parity to the designs, so happy to be led by the team's feels here.

markcipolla:
> Is this the best reason to create a component? Could the design be changed so that we don't have to do this?

Good question! I've made a diagram!

Artboard

To make the spacing (the green bars), the padding, exactly the same as the designs, I had to remove the margin-bottom from the `MessageTileButton`. This meant that `MessageTile`s that _didn't_ have an attribution looked perfect, but the ones with attribution were sitting too close to the button.

 In TC, and the styleguide examples, there wasn't a consistent way of doing the attribution; it was just some random `div`, it wasn't strictly defined, and this way there is _always_ a consistent spacing (the blue bar) between the `MessageTileButton` and the `MessageTileButton`. We can ensure that it exactly matches the design.

With thinking about adding a component, in this case it's really more of a sub-component; it'll never be used outside of the `MessageTile` (hence being in the same folder), and follows the pattern like the other `MessageTile` sub-components: 
- `MessageTileAttribution`
- `MessageTileBody`
- `MessageTileButton`
- `MessageTileHeader`

ZoeJazz:
Ah ok. Thanks for clarifying @markcipolla. I don't know enough about _building_ component libraries to understand other ways to handle this in the codebase, but looks like you've given it plenty of thought. Nice one.

markcipolla:
> Thanks for clarifying @markcipolla.

A good chance to me to explain my thinking (and make sure I fully understand it 😁), and share some knowledge 🙏 .

markcipolla:
> Is there any other way we can solve the top margin issue?

Not consistently to suit all cases.

You could:
1. remove the bottom padding on `MessageTile`. Then everything _inside it_ would have to have margin bottom to push down the next thing, and the last would push down the bottom of the `MessageTile`, effectively replacing the padding bottom it originally had. This unfortunately doens't work, as they'd all require the same spacing; if `MessageTile` had 16px padding, and every component internally was 16px margin-bottom, you're golden. The padding on  `MessageTile` is 24px top and bottom/16px left and right, so without knowing _which_ is the last component, you'd end up with one short side.

**Note**
An addendum, in it's current implementation, we always know which is first (the `MessageTileHeader`), but and could reverse the above (everything has margin-top, with the `MessageTileHeader` hacing a margin top of 24px and everything else 16px), but this means we can never have a `MessageTile` without a title... Unlikely, but maybe? 🤔 

2. You could put a CSS sibling selector, with the last avoiding the margin-bottom... like.... 
```
> * + * {
margin-top: 20px
}
```
It's not the declarative functional React Way™️
Also, I think React-generated markup adds too many child elements to do it consistently.

3. Similar to 2, but a `:last-child` selector... again, too many divs in React-generated to know if it'll work everytime.

4. Add some sort of ".last" css class scenario...  Not all `MessageTile`s have attribution, so we can't rely on the `MessageTileButton` being the last element. So, you could put the identifier on the last element on which ever thing is last... but I think that's a lil' ugly (opinion only).

> I'm not sure about the need for the `MessageTileAttribution` component, given that its only purpose is to set a top margin on a `<div>`.

The other two, at least do two things:
- `MessageTileButton` only adds the margin-bottom and `color='primary'`.
- `MessageTileHeader` sets a typography as variant=h3 and margin bottom. 
- `MessageTileBody` sets a typography as variant=subtitle1 and margin bottom. 

My opinion on having it is so that it's in the UI library, it's declared and reusable, and it's _always_ a consistent size.

> If we do need a new component, maybe we could call it a _footer_ rather than an attribution - that feels like a more generic naming to me.

That feels like a better, more generalised name; it's only _currently used_ for attribution (and I can see a conceptual link between the concept of a message having an attribution, or who the message is from), but `MessageTileFooter` also works for me.

ZoeJazz:
> The padding on MessageTile is 24px top and bottom/16px left and right
Is the issue that the design is inconsistent? If so, let's change the design to reduce the number of components. I think that just means that the version of MessageTile with logo gets additional spacing at the top. I can live with that.

> this means we can never have a MessageTile without a title
This is entirely possible. We could define MessageTile by its title and button.

Let's find the path of greatest simplicity. It's an especially positive outcome if less code means restricting the visual design a little. Working with less variants is most welcome for me :)

z

ZoeJazz:
> I'm not sure about the need for the MessageTileAttribution component, given that its only purpose is to set a top margin on a <div>.

I echo these thoughts. 

If we do feel it's essential, I see an opportunity to simplify the naming: Was a similar extra space built for 'prominent' promos? I wonder if the terms 'Attribution' & 'Prominent' will be lost on us for spacing purposes. If top and bottom spacing is a consistent requirement, then perhaps we call it a 'spacer' component. 

Having said that, I'm much more in favour of creating consistent designs. Solve it in the design so you don't have to fuss. Simplicity for the win :)

tiagoamaro:
> If y'all feel this needs to be split up, it very much can - @markcipolla at https://github.com/conversation/ui/pull/112#issuecomment-549198515

Based on the new component's discussion, probably it's better to create a separate PR for it.

> If we do need a new component, maybe we could call it a footer rather than an attribution - @nullobject at https://github.com/conversation/ui/pull/112#pullrequestreview-310884960

First, agreed on renaming the component to "footer". We can have message tiles with footers, which are not necessarily "attributions".

Given the affirmative above, agreed with @markcipolla on creating a new component for this, as it seems to be a simple question of adding a margin, but tomorrow the "message tile footer" may have different style, font-size, etc. 

IMHO, this should be the responsibility of the UI package.

> If we do feel it's essential, I see an opportunity to simplify the naming - @ZoeJazz at https://github.com/conversation/ui/pull/112#issuecomment-549312093

:point_up: another vote to rename it to "footer" :clap:

ZoeJazz:
> another vote to rename it to "footer"

Would it only ever appear at the foot of a component?

markcipolla:
If we change the design as zoejazz so there's 16px padding all the way
around, we can probably simplify the implementation quite a bit. I'll have
a look on Thursday 😁

ZoeJazz:
> If we change the design as zoejazz so there's 16px padding all the way
around, we can probably simplify the implementation quite a bit. I'll have
a look on Thursday 😁

Yep :) Except can we make it 24, 16 per below?

> The padding on MessageTile is 24px top and bottom/16px left and right

markcipolla:
If we do that, less so. The consistency of the internal margins and the
external padding is the secret sauce 👌. Lets have a chat on Thursday and
brainstorm!

ZoeJazz:
> The consistency of the internal margins and the external padding is the secret sauce

100% agree. May need a bit of help identifying what's not consistent with the current designs though. Chat soon. Enjoy your days off.

markcipolla:
> what's not consistent with the current designs

Ah! I worded myself poorly; the designs are great 🙌  and very consistent. (sent from my phone while waiting 20mins at a cafe fora takeaway because oif the stupid melbourne cup!)

More if we changed the design _slightly_ so the outer padding (24px/16px) was 16px all around, then the outside padding matches the internal margins.

Green is 24px padding, blue is 16px padding, _light_ blue is 16px _margin_.

Left is current (well, you get the idea), and middle is if everything is 16px (I know it's better to give the top and bottom a little more breathing room, but bear with me...)

Artboard

The right takes a little imagination, but each grey bit is a sub-component (the title/body/button), and each has the 16px margin-bottom. We don't need to know which is the last one, as they'll push down the space.

So then we're open to have a bunch of combinations: title/button, or title/body/button, or title/body/button/attribution, or maybe even more, or any combination, and they're all interchangeable.

allconsistentspacings

Anyways, I wanted to say your design is great, and more that CSS can be a nest of spiders. 🕷 

Back to coffee, and my day off! 🎉

markcipolla:
And praise be the powers of caffeine! ☕️ 

Just had a brainwave where we can simplify it, get consistent spacing of any sub-component, _and keep_ the 24px/16px.

Because we're keeping to consistent multiples of 4px units, we change the `MessageTile`'s padding to 24px/8px, and then the last component's margin-bottom (all of them would be 16px) would push down 16px and the outer component pads the bottom 8px (making 24px).

🎉

ZoeJazz:
Heads up - Have updated the MessageTile to consistently have 24px top and bottom across all variations of the design. All this meant was updating the logo variant to have an additional 8px at the top. 

Unfortunately, gallery.io doesn't display masks so it's not very clear in the design :( https://gallery.io/projects/MCHbtQVoQ2HCZWiQBR82z4LY/files/MCEJu8Y2hyDScfjXeeKP3gqEElw_0mfRSGY

Boo to gallery not displaying image Masks properly
Actual margin

ZoeJazz:
Nit: May we please update the title of this PR so it's a little clearer?

From: "Change MessageTile's spacing to get parity with Zoe's designs"
To: "Update MessageTile's padding to 24px, 16px across all variants"

markcipolla:
> Nit: May we please update the title of this PR so it's a little clearer?

On the back of this morning's meeting, I'm going to close this PR and reopen a series of smaller ones :)

And yes, I'll have a more (imperial language? is that the name/phrase @benmacleod?) succint / doing-word title!

ZoeJazz:
> On the back of this morning's meeting, I'm going to close this PR and reopen a series of smaller ones :)

Amazing! Thank you :)

ZoeJazz:
>imperial language?

I found this... 
**imperative-language.** Noun. (plural imperative languages) (computing) Any programming language that consists (mostly) of a series of commands, typically assigning values to objects; a procedural language.

PR #113, opened 2019-11-04
[RFC] Extract duplicate theme selection code to ThemeSelector TCUI component

markcipolla:
So: I'd love your thoughts/opinions/feelings on this. I'm not 100% about the approach, but I've introduced a pattern on TC that's used in a few places, and think there's a component waiting to be extracted.

On https://github.com/conversation/tc/pull/9348, @tiagoamaro wisely [asks about why I've not set up an enum](https://github.com/conversation/tc/pull/9348#discussion_r341586682) of required options for the theme in `PropTypes`.

The reason is because some uses of `EndOfArticle` won't have a `theme` prop; the ones built when the `EndOfArticle` used a `ArticleDonationBanner` instead on a `MessageTile`didn't (as they were super focused/hardcoded on donations). In reworking `EndOfArticle`  tobe more generalised, I have to support _both_ structures of Promo data; some with `theme`, some without.

The pattern that's emerged is a `switch` statement.

```
  let selectedTheme;
  switch (theme) {
    case "core":
      selectedTheme = coreTheme();
      break;
    case "accent":
      selectedTheme = accentTheme();
      break;
    default:
      selectedTheme = defaultTheme();
  }
```

It's duplicated in the (now also generalised) `DesktopBanner`, `EndOfArticle`, `MobileBanner` and `StickyPopup`

The `<ThemeSelector>` is a bridge between a string (the name of a theme), and the theme object itself.

It's job is to know about the themes, alleviating consuming apps, (like TC), from having to know about the themes. Just the _name_ of the theme they want to use. 

I think this is especially useful in the case of Promos, where we're storing the theme as a string.

It's to be used in place of a `ThemeProvider`, as the `ThemeSelector` wraps a `ThemeProvider` under the hood.

It'll also work moving forward with SSR, allowing us to pass down the other `ThemeProvider` props: `sheetsRegistry` & `generateClassName`

------------

To check it out, open up Storybook, and http://localhost:9001/?path=/story/themeselector--overview


ZoeJazz:
This looks really promising @markcipolla :) Nice work.

markcipolla:
> This looks really promising @markcipolla :) Nice work.

🙇 🙌

src/ThemeSelector/ThemeSelector.jsx
```diff
@@ -0,0 +1,81 @@
...
+
+ThemeSelector.propTypes = {
+  /**
+   * The child components to which the theme will be applied.
+   */
+  children: PropTypes.node.isRequired,
+  /**
+   * The allowed theme choices.
+   */
+  theme: PropTypes.oneOf(['accent', 'core', 'default'])
```
tiagoamaro: :heart:!

Lovely! If we mistakenly use a unregistered theme, it'll only pop out errors on our consoles, but it'll fallback to the default! Nice! :clap:

src/ThemeSelector/ThemeSelector.test.jsx
```diff
@@ -0,0 +1,41 @@
...
+        expect(wrapper.find(ThemeProvider).props().theme).toBe(defaultTheme())
+      })
+    })
+
+    describe('accepts no value', () => {
+      it('passes the default theme to the ThemeProvider component', () => {
+        const wrapper = mount(<ThemeSelector theme='default'>Hello</ThemeSelector>)
+        expect(wrapper.find(ThemeProvider).props().theme).toBe(defaultTheme())
+      })
+    })
```
tiagoamaro: Typo: forgot to remove the `theme='default'` prop from the `ThemeSelector` on test scenario.

markcipolla: AH! 🙌 Good spotting!

Fixed.

src/ThemeSelector/ThemeSelector.jsx
```diff
@@ -0,0 +1,81 @@
...
+}
+
+export default ThemeSelector
```
tiagoamaro: Nit: instead of `const Component = (props) {}`, we could use functions: 

```js
export default function Component(props) {}
```

markcipolla: I'll make up teeny tiny follow up PR to make this change @tiagoamaro

markcipolla: Changed in https://github.com/conversation/ui/pull/116

PR #114, opened 2019-11-07
Change MessageTile padding to 24px 16px 16px 16px

markcipolla:
https://trello.com/c/fw5FrrXi/1330-change-messagetiles-spacing-to-top-bottom-24px-and-left-right-16px

*Why*

We want overall spacing of the `MessageTile` to be 24px/16px. We need to factor in that internal components have their own margin (to space themselves appropriately), and need to factor that in.

*How*
We use consistent multiples of 4px as base units, so by changing the
`MessageTile`'s padding top/bottom to 24px/16px, the last component's margin-bottom (all of them are 8px) would push down 8px, and the outer component pads the bottom 16px (making 24px... 8px + 16px = 24px, matching the padding-top).

Artboard

Left is master, right is this PR (note the _slight_ change to the spacing on all `MessageTile` examples.
Artboard

**Also note** that the one with Misha's attribution isn't quite right (it's not the same size as the others). I'll look at implenting a `MessageTileFooter` component to wrap up whatever would go there, and give it margin as well like the other `MessageTile` sub-components. The discussion is here (https://github.com/conversation/ui/pull/112#issuecomment-549319190), and that'll be a separate PR.

### DIFF
--- a/src/MessageTile/MessageTile.jsx
+++ b/src/MessageTile/MessageTile.jsx
@@ -9,7 +9,10 @@ const styles = theme => ({
   box: {
     borderRadius: theme.spacing(0.5),
     marginBottom: theme.spacing(2),
-    padding: theme.spacing(2)
+    paddingTop: theme.spacing(3),
+    paddingBottom: theme.spacing(2),
+    paddingLeft: theme.spacing(2),
+    paddingRight: theme.spacing(2)
   }
 })
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * Update docs in example
 * Change MessageTileHeader from h3 to h5
+* Minor padding change to MessageTile
 
 # 2.16.0 (2019-10-09)
 


nickbrowne:
The before and after screenshots don't seem any different to me

markcipolla:
> The before and after screenshots don't seem any different to me

🙏 

Ah! It hot reloaded in between switching to master and screenshotting. Excellent pickup!
I've updated the image in the description.

ZoeJazz:
> Also note that the one with Misha's attribution isn't quite right (it's not the same size as the others). I'll look at implenting a MessageTileFooter component to wrap up whatever would go there,

Why will this be different to the others?

markcipolla:
> Why will this be different to the others?

The `div` around the attribution doesn't have any margin-bottom.
I'm adding a `MessageTileFooter` component to replace the ad-hoc `div` we have in the examples and in TC. It'll have the margin-bottom, so it'd push down the spacing in the same way as the `MessageTileButton`

Adding in https://github.com/conversation/ui/pull/115

ZoeJazz:
> The div around the attribution doesn't have any margin-bottom.

Ah ok, just to clarify how the design system is rolling out - Is this what we're aiming to do?
1. Apply consistent padding for all containers (ie MessageTile) ✅_Donesies_
2. Apply 0px margin around all sub-components (ie button, image, avatar, text, etc) 
3. Create a set of spacers (ie 4, 8 & 16px) to consistently use between all sub-components

markcipolla:
> 2. Apply 0px margin around all sub-components (ie button, image, avatar, text, etc)
> 3. Create a set of spacers (ie 4, 8 & 16px) to consistently use between all sub-components

No? Making spacers? I don't understand. Can you explain that bit again?

markcipolla:
> Can you explain that bit again?

Succinctly described in https://medium.com/fed-or-dead/handling-spacing-in-a-ui-component-library-70f3b22ec89.

Worth a read! We should discuss this more

ZoeJazz:
> No? Making spacers? I don't understand. Can you explain that bit again?

Ah sorry. I was just trying to clarify the big picture here but I don't think I quite grasped it in my previous comment. 

My two cents: I like how pragmatic this PR is but I feel like this solution will result in further 'clean up' work down the track. Is there a simpler, more adaptive solution that avoids further specificities in our styling?

markcipolla:
> Ah sorry. I was just trying to clarify the big picture here but I don't think I quite grasped it in my previous comment.

All good! 👍 for shared understanding; definitely helped me understand both this work and where we can take the UI library moving forward.

> My two cents: I like how pragmatic this PR is but I feel like this solution will result in further 'clean up' work down the track. Is there a simpler, more adaptive solution that avoids further specificities in our styling?

Good question. There is some uncomponentised things in this component as it is, so this locks it down so it a) is built and documented in the UI library, and b) brings it inline with the designs.

You've raised excellent questions about vertical alignment _as a whole_ across the library, so that definitely needs some buy-in across the team as a structure we'd use in _all_ components.

I don't see a way we can get what we need now, which is migrating the TC components to using a shared `MessageTile`, without making _a_ change, and this is the smallest set of changes to get parity with the designs I can think of.

If we want to change _how_ we do vertical spacing, we'll have to make changes (and have cleanups), both to this component (and subcomponents), but eventually _all_ components as well.

ZoeJazz:
Ok. We don't have stakeholders breathing down our necks with component library work. So, we have the luxury to not just build assets that look right, but also apply adaptive, unifying, simple solutions that make our lives easier in the long run.

If this means going back to the drawing board for this particular PR, then that could be a great result if it avoids the need to refactor later on. 

I'm conscious that I'm not deep in the code, so perhaps there's something I'm missing here. I wonder if it's best discussed with @jameshill @nickbrowne or @nullobject ? Whatever we choose to do, I'll update the sketch library to suit.

src/MessageTile/MessageTile.jsx
```diff
@@ -9,7 +9,10 @@ const styles = theme => ({
   box: {
     borderRadius: theme.spacing(0.5),
     marginBottom: theme.spacing(2),
```
nullobject: I'm not following 100%.

Should the `marginBottom` spacing be changed to `theme.spacing(1)` (i.e. 8px)? If the bottom padding is 2 units, and the bottom margin is 1 unit, then that adds up to 3 units (i.e. 24px).

markcipolla: `MessageTile`'s margin is spacing _outside_ the containing box, so it's for spacing between it and other components. 

I do believe, however, that I've got my units mixed up. The code is right, but not the title. Apologies for the confusion.

The padding on the `MessageTile` is 24px 16px 16px 16px. We're wanting spacing of 24px 16px overall.

With the `MessageTile`'s padding-bottom of 16px, and the `MessageTileButton`/`MessageTileFooter` having a margin-bottom of 8px, there's the 24px we need to get a _spacing_ of 24px 16px.

markcipolla: I've amended the title.

nullobject:
Oops, sorry I deleted my comment.

I think I understand that the inner button has it's own 8px bottom margin, which when added to the bottom padding of the container, is 24px.

But what if we have a message tile without an inner button. Is that going to throw things out?

markcipolla:
> I think I understand that the inner button has it's own 8px bottom margin, which when added to the bottom padding of the container, is 24px.

Yes. Correct.

> Oops, sorry I deleted my comment.

All good! 🙇 🙌 
I was replying, and then github let me click the button and nothing happened and I was confused for a tick!

> But what if we have a message tile without an inner button. Is that going to throw things out?

Yes/sort of. If we have a message tile without a button _or_ a footer (it'd have to have at least one or the other, I guess), then yes. 

This is because the `MessageTileBody` has inconsistent margin compared to thee other sub-components (of 16px, rather than 8px). 

That introduces an edge case (where the `MessageTileBody` is the _last_ sub-component inside `MessageTile`) where it'd have a smidge too much space ... if we had a message tile with a body, but no button nor footer, for example.

All _current_ examples have a button.

As it currently stands, we could mix-and-match to have (examples from http://styleguide.theconversation.com/?path=/story/messagetile--examples): 
- a title, body and a button (the republish panel / donation CTA / newsletter subscription)
- a title, body, button and a footer (end of article donation CTA)

We could also *optionally* have:
- a title and a button (no case for this yet)
- a title, body and footer (no case for this yet)
- a title and a button (no case for this yet)
- a body and a button (no case for this yet)

And these cases where I can't see a use-case yet:
- a body and a footer (no case for this yet, and to be honest, can't see a need)
- a title and a footer (no case for this yet, and to be honest, can't see a need)
- a button and a footer (no case for this yet, and to be honest, can't see a need)

Basically, the only one we can't have is where the body is the last component.

It's a teeny bit moot, as they're cases we don't have any uses for (granted: yet), though definitely worth considering moving forward with how we want to structure vertical spacing, as we don't have any cases where we're trying to make a message tile without a button _nor_ a footer.

markcipolla:
Closing for same reason as https://github.com/conversation/ui/pull/115

PR #115, opened 2019-11-07
Create a MessageTileFooter component

markcipolla:
*Why*

We have an ad-hoc `div` in the storybook examples and in TC's `EndOfArticle`, with no styling.
To make it like the `MessageTileButton` component and have the same consistent margin-bottom to push down the bottom of it's parent (explained in https://github.com/conversation/ui/pull/114), I've made a new component to sit in this spot.

Left is master (with unstyled divs), right is this PR (using the sub-component)
Artboard

Following on from discussion at https://github.com/conversation/ui/pull/112#issuecomment-549319190

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * Update docs in example
 * Change MessageTileHeader from h3 to h5
+* Add MessageTileFooter sub-component
 
 # 2.16.0 (2019-10-09)
 
--- a/src/MessageTile/MessageTile.jsx
+++ b/src/MessageTile/MessageTile.jsx
@@ -19,24 +19,43 @@ const styles = theme => ({
  * call-to-action button at the bottom.
  *
  * ```js
- * import { MessageTile, MessageTileHeader, MessageTileBody, MessageTileButton } from '@theconversation/ui'
+ * import {
+ *   MessageTile,
+ *   MessageTileHeader,
+ *   MessageTileBody,
+ *   MessageTileButton,
+ *   MessageTileFooter,
+ *   Avatar,
+ *   Person,
+ *   Typography
+ * } from '@theconversation/ui'
  *
  * <MessageTile>
- *  <MessageTileHeader>
- *    <img style={{ width: 26, marginRight: 6, marginBottom: 8 }} src={by} />
- *    <img style={{ width: 26, marginRight: 6, marginBottom: 8 }} src={cc} />
- *    <img style={{ width: 26, marginBottom: 8 }} src={nd} />
- *    <div>We believe in the free flow of information.</div>
- *  </MessageTileHeader>
+ *   <MessageTileHeader>
+ *     Before you go...
+ *   </MessageTileHeader>
  *
- *  <MessageTileBody>
- *    All our articles can be republished for free, online or in print, under the Creative Commons licence.
- *  </MessageTileBody>
-
- *  <MessageTileButton onClick={action('clicked')}>
- *    Republish for free
- *  </MessageTileButton>
- *</MessageTile>
+ *   <MessageTileBody>
+ *     <Typography variant='body1' paragraph>
+ *       It is easier than ever before for vested interests to spread
+ *       disinformation on vital matters of public interest. If you want
+ *       to know what's really going you need to hear from the experts
+ *       willing to drill down to the truth. But we can't do that vital
+ *       work unless readers donate. Please make a donation.
+ *     </Typography>
+ *   </MessageTileBody>
+ *
+ *   <MessageTileButton fullWidth={false} prominent onClick={action('clicked')}>
+ *     Donate now
+ *   </MessageTileButton>
+ *
+ *    <MessageTileFooter>
+ *      <Typography gutterBottom>
+ *        <Person name='Misha Ketchell' caption='Editor' />
+ *      </Typography>
+ *      <Avatar size={48} src='https://cdn.theconversation.com/avatars/13/width238/image-20181205-186070-f58pk2.jpg' />
+ *    </MessageTileFooter>
+ *  </MessageTile>
  * ```
  *
  * You can change the internal styling by wrapping the whole thing in a `ThemeProvider`
--- a/src/MessageTile/MessageTileFooter.jsx
+++ b/src/MessageTile/MessageTileFooter.jsx
@@ -1,0 +1,30 @@
+import PropTypes from 'prop-types'
+import React from 'react'
+import withStyles from '@material-ui/core/styles/withStyles'
+
+const styles = theme => ({
+  footer: {
+    marginBottom: theme.spacing(1)
+  }
+})
+
+export const MessageTileFooter = ({ children, classes }) => {
+  return (
+    <div className={classes.footer} key='footer'>
+      {children}
+    </div>
+  )
+}
+
+MessageTileFooter.defaultTypes = {
+  children: {}
+}
+
+MessageTileFooter.propTypes = {
+  /**
+   * The nested components that go in the body
+   */
+  children: PropTypes.node.isRequired
+}
+
+export default withStyles(styles)(MessageTileFooter)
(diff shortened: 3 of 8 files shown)


markcipolla:
> Creating a component called 'MessageTileFooter' doesn't sound like it's going to scale well as it's intrinsically linked to only one component type. I'd feel more on board with this approach if it benefited more than just the MessageTile component.
> 
> @nullobject raised some good points in PR #112. Is there any other way we can solve the margin issue?

We have a `MessageTileHeader`, a `MessageTileBody`, a `MessageTileButton`, and this codifies what we already have (some ad-hoc code in two codebases) as an estabhlished pattern. We can put whatever footer we want in there, if at all, as it's optional.

markcipolla:
> @nullobject raised some good points in PR #112. Is there any other way we can solve the margin issue?

Then we're back to the discussion on https://github.com/conversation/ui/pull/112, and not without losing something; the 24px/16px spacing we want means we need to do something.

Basically, it's just a box around either this:
Screen Shot 2019-11-07 at 3 53 37 pm

But all it does is be a box that has a pre-defined margin (unlike having random divs in other parts of the codebase that might make whatever margins they want), so it works with the other boxes we already have (MessageTileHeader, a MessageTileBody, a MessageTileButton). 
These sub-components are never used anywhere but from inside a `MessageTile`, and are needed to execute on the variants of the MessageTile.

If we have a problem more generally with sub-components, then sure, that's a bigger discussion (which I'm totally on-board with, these are super good questions 🙌 ) about structuring the UI library and code.

ZoeJazz:
> We have a MessageTileHeader, a MessageTileBody, a MessageTileButton,

Ah I see. Apologies, my mistake. Where possible, structuring components in generic terms ie Header5, Body2, Button-small, etc would be a huge win. That's how I've set it up in Sketch to avoid ad-hoc variants. Visually, the output looks great @markcipolla. I'm just wondering if there's other ways to solve this that benefits other parts of the component library.

benmacleod:
@ZoeJazz are you still requesting changes?

benmacleod:
@markcipolla should the examples start referring to `MessageTileFooter` now? https://github.com/conversation/ui/blob/fc47f12a75f02f08b2cc07409554ee9268f28948/src/MessageTile/MessageTile.jsx#L35

markcipolla:
> @markcipolla should the examples start referring to `MessageTileFooter` now?

I'd only added it for the examples that had a use for one, as it's an optional component only in one case, like in https://github.com/conversation/ui/pull/115/files#diff-a44cec5807168d1ed38cc140bec962ceR105-R110

We could definitely show more examples, and ones that are move obvious to it's use.

benmacleod:
> We could definitely show more examples, and ones that are move obvious to it's use.

So, are you saying you will update the examples? They don't all have to use a footer, but it seems to me like, as an end user of this library, I'd find it weird to have this component in there without any examples of how it's used.

nullobject:
Is the idea to get rid of the `MessageTileButton` once this is merged?

markcipolla:
> Is the idea to get rid of the `MessageTileButton` once this is merged?

Nope. It's a sibling to the `MessageTileButton`. In the same way it's a sibling to the title / body.

I mean, the entire vertical-spacing could do with more thinking and work across _all_ components, and the article @ZoeJazz suggested has some interesting ways to solve it (and do away with all the sub-components 🎉 , which I'm a fan of, but that would have to be part of a larger discussion on vertical spacing and how we want to do it), but at the moment, no.

markcipolla:
> So, are you saying you will update the examples?

Sure. I've change the default example (on the MessageTile Overview) to show the example (the end of article donation call-to-action) that uses all the components.

Screen Shot 2019-11-08 at 4 59 13 pm

nickbrowne:
One of the concerns in the previous PR seemed to be that we had a component just to provide a layout (the button), but this seems to be doing much the same thing.

What is it that this new component actually lets us achieve that just having the padding on the outer component won't?

Having tried to get my head around what's going on with these pull requests I think we need to spend a bit more time thinking about how we handle spacing and layout. I think adding specialised components to provide layouts should be reserved for really special cases (eg. avatars mid dialog border).

Layouting composite components should be as simple as slapping things together without thinking about it too much (this is how MUI works out of the box). If that's not happening for us then we've made a mistake at some point and we need to work out what that mistake is.

nullobject:
> Nope. It's a sibling to the MessageTileButton. In the same way it's a sibling to the title / body.

I think the `MessageTileButton` should have a good reason for being around. IMHO a padding tweak doesn't justify its existence any more.

If we look at what MUI does with their `Card` and `Dialog` components (https://material-ui.com/api/card-actions/), they have an `Actions` sub-component which is somewhat akin to this new footer component.

It has a uniform top and bottom padding, the `Card` itself also has uniform padding.

Screenshot from 2019-11-08 17-05-32

Importantly, they don't define a `CardButton` or `DialogButton`, because it is too specific to justify adding it to the library.

markcipolla:
> What is it that this new component actually lets us achieve that just having the padding on the outer component won't?

Yeah, it's pretty confusing.
The design has 24px 16px as the padding.
Internal sub-components need spacing between each other. So each component has a margin.
But the last component pushes down the bottom of the parent, making it not the right size.

This component _only_ codifies (by putting it in the ui library) what we already have, instead of random divs reimplemented in different places. This locks down the design to exactly as specified, making sure the spacing is consistent. We already have this "component" in TC, but it's not a component; it's literally some div somewhere. The design and use-cases _already have this shown_, but not as a component.

It adds to the ugliness, sure, by adding another sub-component it makes it consistent, it follows the established pattern, and then we can refactor it. I don't like this. This component shouldn't exist.

But we make small incremental changes to a codebase. Moving it towards what we want.

End of the day:

a) this is to fix a 8px spacing bug. 
Screen Shot 2019-11-08 at 5 07 37 pm
b) It makes a component (exactly like the other ones) so the design is documented from _non-componentised code_.
c) I'm super not wedded to this. This will not be the hill I die on :) 

But if we have an example in the UI library that shows something, but it's not a component, that feels far worse.

I agree we should look at how we want to do vertical spacing, and we can do it better.

markcipolla:
> I think the `MessageTileButton` should have a good reason for being around. IMHO a padding tweak doesn't justify its existence any more.

I agree 100%.

All the `MessageTile` sub-components **are exactly only for having a _margin-_tweak**. That's all they do. That's it.

I'd still rather make one like the others, so _it's documented_, and then changing to not need them at all.

markcipolla:
> What is it that this new component actually lets us achieve that just having the padding on the outer component won't?

Functionally, nothing.
It componentises and documents undocumented and uncomponetised code.

 
> Having tried to get my head around what's going on with these pull requests I think we need to spend a bit more time thinking about how we handle spacing and layout. I think adding specialised components to provide layouts should be reserved for really special cases (eg. avatars mid dialog border).

Also agree.

markcipolla:
Originally, six months ago, we added the sub-components to `MessageTile` make it more consistent with _existing_ components, like the `Dialog` component.

Screen Shot 2019-11-08 at 6 07 53 pm

This [commit](https://github.com/conversation/ui/commit/e0b1ff00d5ed796bc746367f2e790b797cab39f1) from this [PR](https://github.com/conversation/ui/pull/82) detailed the reasons why.
And lots have changed between then and now. We've learnt much more.

> I'd like to propose that we find a different solution.

In no way am I disagreeing. I wholeheartedly agree that we need to discuss vertical spacing.

We have undocumented componentry in the component library. 
That feels like a real code-smell in a component library.

This addresses that, bringing it to light that there's a larger problem we'd like to address. 

It brings parity with the design. 

It unblocks changes for TC, which in turn, unblock the _reuse of shared components_: using the `MessageTile` for both the `RepublishPanel` and the `EndOfArticle` CTA.

That felt like a stepping stone; a small, incremental change, addressing a problem we can easily address first, throwing light (and good discussion) to highlight the ugliness (that vertical spacing to handle any/all cases is not great, across _all_ components), and then solve it, after discussions _as a team_. Much like @nickbrowne's previous suggestions (at points in the codebases lifetime, TC I think) that duplication is a handy tool to highlight issues in the codebase. 

Through seeing the issues then can you understand the problem, and better understand the solution.

Let's have a team chat about how to better handle this next week.

PR #117, opened 2019-11-08
Add resets to heading typography variants

nickbrowne:
This isn't ideal, but it was far harder than expected to quarantine TC's styles. (biggest blocker and bordering on impossible to change at the moment is the content pipeline). This ended up being the easiest way to prevent TC's legacy CSS from messing with MUI typography heading variants.

Initially I planned to reset every property that we set in TC (like the trello card mentions), but that caused additional problems due to class specificity and the way MUI mixes classes together to build the styles for Typography components.

In the end, I experimented until I only had resets for just the properties that were likely to cause (and have caused) the most grief while effectively being a no-op visually - so if you are testing locally, you shouldn't see any visual differences in this PR.

This may mean we could encounter problems in less travelled parts of TC, but we can always add more resets if needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Unreleased
 
+* Add stylesheet property resets to heading variants of Typography
+
 # 2.17.0 (2019-11-07)
 
 * Adds a ThemeSelector component
--- a/src/Typography/Typography.jsx
+++ b/src/Typography/Typography.jsx
@@ -15,8 +15,25 @@ import withStyles from '@material-ui/core/styles/withStyles'
  * </Typography>
  * ~~~
  */
+
+// Default overrides which prevent TC's legacy top level styles from having
+// a negative impact on typography components.
+//
+// Note that some properties have been skipped intentionally, such as margin.
+//
+// The margin is set via a different class on a Typography by MUI itself, which
+// is already specific enough to override TC's styles. Including a reset
+// for those properties (margin, color, etc) would cause more harm than good.
+const LEGACY_HEADING_RESETS = {
+  fontSize: 'initial',
+  fontWeight: 'initial',
+  textTransform: 'initial',
+  letterSpacing: 'initial'
+}
+
 const styles = theme => ({
   h1: {
+    ...LEGACY_HEADING_RESETS,
     fontFamily: 'Montserrat',
     fontWeight: 700,
     fontSize: '1.75rem',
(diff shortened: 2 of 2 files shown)


markcipolla:
@nickbrowne So, to test this, would this be `npm link`?

markcipolla:
Are there any places on TC you've seen where it would fix problems?

nickbrowne:
> @nickbrowne So, to test this, would this be `npm link`?

If that works for you, I just did an `npm install` with the absolute path.

> Are there any places on TC you've seen where it would fix problems?

The most obvious ones have already been fixed, but I couldn't continue with that way of fixing things because it became too difficult:

https://github.com/conversation/tc/pull/9275
https://github.com/conversation/tc/pull/9313
https://github.com/conversation/tc/pull/9337

I will likely revert some of those changes after this merges and we cut a new version.

jameshill:
@nickbrowne oh interesting so you're going to revert those PRs and remove the legacy class?

That's a shame.. what ended up being the straw that broke the camels back in content pipeline?

I have no idea the best way to isolate those components, but you couldn't target them by a class on the content body's container... 

we wouldn't be able to have tc/ui inside the content body though, but we're not do that at the moment.

nickbrowne:
> That's a shame.. what ended up being the straw that broke the camels back in content pipeline?

Without going into too much detail, Rohan and I tried to extend the markdown library (according to a blog post from 2013 that only exists on the internet archive) to have more control over how it outputs heading elements. 

But because it's c extensions and metaprogramming, it was far harder than expected. Then when we finally got it working, we ran into problems where we ended up with those classes ended up in unexpected places like feeds and apple news.

We thought about doing some gsubbing, but then I backed away from it all together, I wasn't willing to add more complexity and switches to the content pipeline.

We also tried the more targeted css that you mentioned, but that broke a bunch of other styles because the change in specificity meant that the way the styles cascaded was different. There's also the issue that it would break ui components rendered into a content body even worse, like you mentioned.

Basically all roads led to hell :fire:, so I think the best bet in the short term is what we've got here and then visually unifying the two worlds when we're ready.

jameshill:
Sounds good... I can't test this at the moment, but it gets a 👍 from me... if someone else can have a play with it alongside TC that would be marvellous.

markcipolla:
> If that works for you, I just did an `npm install` with the absolute path.

I've tried `npm install` and `npm link`, and no luck; `npm i` with an absolute path won't build, and `npm link` seems to do nothing (or, more likely, I'm using it wrong).

@nickbrowne Are you able to post up some steps on how to manually test this?

markcipolla:
@nickbrowne did a screen share (the code looked good, but nice to see it in action)

👍

nickbrowne:
> @nickbrowne Are you able to post up some steps on how to manually test this?

Will open a PR to add something to the readme because I also wasted a bunch of time trying to get it to work again today.

PR #120, opened 2019-11-14
Change example of MessageTile to have match how it's implemented on TC

markcipolla:
https://trello.com/c/CfgJGmwb/1421-show-use-of-sizesmall-button-on-republishpanel-examples

**Why**

The MessageTile was designed for text of the length "Republish for free", but implemented on TC with "Republish this article".

This caused some line wrapping, so we decided to use the `small` variant of the `Button`.

This _only_ changes the example in Storybook. The changes are already implemented on TC.

**Before**
Screen Shot 2019-11-14 at 1 35 40 pm

**After**
Screen Shot 2019-11-14 at 1 35 48 pm

### DIFF
--- a/src/MessageTile/stories/examples.jsx
+++ b/src/MessageTile/stories/examples.jsx
@@ -41,8 +41,8 @@ export default () => (
             All our articles can be republished for free, online or in print, under the Creative Commons licence.
           </MessageTileBody>
 
-          <MessageTileButton onClick={action('clicked')}>
-            Republish for free
+          <MessageTileButton size='small' onClick={action('clicked')}>
+            Republish this article
           </MessageTileButton>
         </MessageTile>
       </ThemeProvider>


RohanM:
Looks great and matches what we're doing on TC. I notice that the overview page still uses the `Republish for free` text and the regular font size; would it make more sense to update that to match the example, or keep it as is for simplicity?

tiagoamaro:
@RohanM's right. Also, a nit: `StickyPopup` and `MessageTile` are the only components with an "Overview" and "Examples" stories.

In both cases, should an "overview" be simply a text introduction of the component, and leave for other stories to document its usage?

nullobject:
I set up a vote in Slack to try to pin down a time for this meeting: https://conversation.slack.com/archives/G02EEH7KF/p1574055170130000

PR #121, opened 2019-11-20
Remove now-unused ThinDonationBanner component

markcipolla:
https://trello.com/c/OVHqPRbJ/1467-deprecate-hardcoded-donation-related-promo-components

**Why?**
We had donation-focused hardcoded (in that the colours were hardcoded red, rather than using themes) when we launched promos to get moving quickly

Later, we built themeable components to replace them. In this components case, `ThinBanner`.

We've migrated TC from using the hardcoded banners, and need to remove them from the UI library.


markcipolla:
> NIT: Maybe drop something in the changelog to jog our memory later?

Yes! I'll do that now. Tip top idea.

PR #132, opened 2020-08-31
Add tree shaking support

nickbrowne:
Right now when using any single component from this library, you get everything but the kitchen sink whether you like it or not, which isn't ideal.

With a couple of tweaks we can let webpack at least discard parts that don't actually get used with [tree shaking](https://webpack.js.org/guides/tree-shaking/).

In the following screenshots you can see (we've got a lot of work to do... but anyway) that when part of the ui library doesn't get used, it drops out of the bundle automatically. In this example, it's what happens when you don't use the datepicker.

Screenshot at 2020-08-31 16-45-07

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ node_modules:
 	@npm install
 
 dist:
-	@npx babel --copy-files --out-dir dist src
+	@npm run build
 
 test:
 	@npx jest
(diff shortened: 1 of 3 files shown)


nickbrowne:
Notes for upgrade, in TC where we pull in a folder of logos in a `require.context`, will have to change it to: 
`"@theconversation/ui/dist/esm/logos"`

tiagoamaro:
About https://github.com/conversation/ui/pull/132#issuecomment-683597386: might be interesting adding this to the CHANGELOG when releasing a new version of the `@conversation/ui` package.

PR #136, opened 2020-10-14
Max characters client side validation

tiagoamaro:
**Why?**

We want to have a visual feedback of the character limit of a text field

**How?**

State management happens inside the `TextField` component, being flexible enough to work when:

- Passing the `defaultValue` or `value` prop (counting initial values)
- Input with a `helperText` prop
- `multiline` inputs

**Manual testing steps**

- Checkout this branch
- Run the storybook (e.g.: `make storybook`)
- Visit the `TextField` states story (e.g.: http://localhost:9001/?path=/story/text-fields--states)
- You should see the new "Maximum length validation" input in the "Example" field

**Screencast**

max-length-validation-example

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Unreleased
 
+* Allow text fields to have max length validation with character count
+
 # 2.22.0 (2020-10-02)
 
 * Allow passing custom props to MUI's Dialog on `DonationDialog`
--- a/src/TextField/TextField.jsx
+++ b/src/TextField/TextField.jsx
@@ -67,6 +67,7 @@ export const TextField = ({
   label,
   required,
   maxLength,
+  maxLengthCountText,
   ...other
 }) => {
   const [errorState, setErrorState] = useState(error)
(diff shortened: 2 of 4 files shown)


ZoeJazz:
Wonderful :) Thank you @tiagoamaro 

What happens if the helper text is especially verbose? See screenshots of what I had in mind... 
Screen Shot 2020-10-15 at 9 52 55 am
Screen Shot 2020-10-15 at 9 53 46 am

tiagoamaro:
@ZoeJazz, right now the input is adjusting its size based on the helper text's width, which is causing a curious "long cat" effect :smile_cat::

<details>

- Desktop

image

- Mobile

image

</details>

Will fix it up for this case!

tiagoamaro:
@ZoeJazz, updated the helper text grid to match https://github.com/conversation/ui/pull/136#issuecomment-708702649. Here's how would work:

- On desktop with unconstrained text field size, we'll have the "long cat" effect

image

- On desktop with constrained text field size, we'll break words as expected

image

- On mobile

image

ZoeJazz:
Perfection! Thank you @tiagoamaro

nickbrowne:
I'm just taking a look at this now.

nickbrowne (approved):
@tiagoamaro can you just sanity check my changes in https://github.com/conversation/ui/pull/136/commits/797e3254e4737f4ed2b426814819e53b3ddab5a1? If you're happy go ahead and merge. I did my best to simplify the custom styles and explain what was going on with the existing ones in comments.

tiagoamaro:
Awesome @nickbrowne, confirmed that https://github.com/conversation/ui/pull/136#pullrequestreview-510199073 change looks great! Thank you very much!

PR #140, opened 2020-12-03
use native select for Dropdown component

tiagoamaro:
**Why?**

MUI custom select creates too much friction for users and test utilities. Let's use the native version instead.

If we want to use it anyway, we can disable the native version by passing the `<Dropdown native={false} />` prop.

**How to test  it?**

- Checkout this branch
- Run storybook (e.g.: `make storybook`)
- Read the new documentation on "Dropdown" overview (e.g.: http://localhost:9001/?path=/story/dropdowns--overview)
- See the native select in action checking the "Dropdown" "States" stories (e.g.: http://localhost:9001/?path=/story/dropdowns--states)

### DIFF
--- a/src/Dropdown/Dropdown.jsx
+++ b/src/Dropdown/Dropdown.jsx
@@ -39,13 +39,10 @@ const styles = theme => {
 }
 
 /**
- * The `<Dropdown>` component allows users to select from a list of items,
- * rendering MUI's native `Select` by default. Keep in mind that non-native
- * `Select` components only allow `MenuItem` components as children.
+ * The `<Dropdown>` component allows users to select from a list of items.
  *
  * ```js
- * // Native Dropdown
- * import { MenuItem, Dropdown } from '@theconversation/ui'
+ * import { Dropdown } from '@theconversation/ui'
  *
  * <Dropdown
  *   helperText='Select a colour'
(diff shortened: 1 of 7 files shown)


src/Dropdown/Dropdown.jsx
```diff
@@ -39,15 +39,34 @@ const styles = theme => {
 }
 
 /**
- * The `<Dropdown>` component allows users to select from a list of items.
+ * The `<Dropdown>` component allows users to select from a list of items,
+ * , rendering MUI's native `Select` by default. Keep in mind that non-native
```
benmacleod, line 43:
```suggestion
 * rendering MUI's native `Select` by default. Keep in mind that non-native
```

benmacleod (changes requested):
Looks good, and it's a good idea to allow the native implementation, especially since we're already using it in the wild in donations.

Is there actually still a use case for the non-native implementation? Given that it makes for complex testing maybe without benefits, should we actually get rid of non-native dropdowns? Would that break anything if this library suddenly switched?

If we do continue offering non-native dropdowns, we should probably show them in the `Dropdown` stories. Maybe as a variant on a "variants" page like `src/Button/stories/variants.jsx`?

tiagoamaro:
@benmacleod, I think it's unlikely we'll use the non-native variant, but documenting that it's available as it's a simple replication of MUI's `Select` component is a really nice idea!

Liked the `stories/variants.jsx` approach. Will add it to the PR!

benmacleod (approved):

benmacleod (approved):
If we are definitely going to allow non-native `Dropdown`s (which seems fine to me), then I think this is the best way to expose and document them 👍🏻

PR #144, opened 2021-02-17
Create the SelectMenu and SelectMenuItem components

tiagoamaro:
This component renders a list of options for users to selected,  allowing the developer to control which item is selected and
what actions are going to be triggered after an option is clicked.

**Screencast**

select-menu


ZoeJazz:
Looking good @tiagoamaro :) 

Only thought is that the menu should close after a selection is made, like this:
https://user-images.githubusercontent.com/7854383/108271934-a6734400-71c5-11eb-9d42-0f80133ef1ad.mov

tiagoamaro:
Adjustments:

* Menu should close after an option is selected
* Hovering an item will darken its background
* A selected item won't have a visual feedback that's selected

ZoeJazz:
Slight tweak on these adjustment notes:

- Menu should close after an option is selected
- Hovering an item will darken its background
- A selected item will have a checked icon as feedback that's selected

tiagoamaro:
@ZoeJazz , made all adjustments:

* Menu closing after an option is selected: this behaviour is now present on the "Usage" example in the storybook. This behaviour is completely controllable for developers, so if we want to leave the `SelectMenu` opened after clicking an option, we can
* Hovering an item will darken its background: done!
* Selected item with a checked icon: also, something that'll vary each `SelectMenu` implementation. Added the checked icon to the selected menu item in the usage example as well!

src/SelectMenu/SelectMenu.jsx
```diff
@@ -0,0 +1,70 @@
...
+    </Popover>
+  )
+}
+
+SelectMenu.defaultProps = {
+  anchorEl: undefined
+}
+
+SelectMenu.propTypes = {
+  /**
```
benmacleod, line 51: Would it make sense to add some demonstrations of the effects of changing these props (`anchorEl`, `onClose`, `open`) in the Usage stories?

tiagoamaro: @benmacleod, `anchorEl`, `onClose` and `open` are required for the select menu to function. I added a couple of actions, so readers can better understand when those events and props are changed and called:

example

ZoeJazz:
Perfection :)

benmacleod (approved):
OK, I get it now. Thanks for the explanation and the extra elucidation in the story 👍

PR #145, opened 2021-02-18
Bump versions: material UI core (4.5.0 -> 4.5.2) and labs (4.0.0-alpha23 -> 4.0.0-alpha30)

tiagoamaro:
**Why?**

MUI introduced their [`Autocomplete` component](https://material-ui.com/components/autocomplete/) on version 4.5.2 and labs 4.0.0-alpha30.

We'll want to use that in the future, so let's update both packages.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1447,37 +1447,25 @@
       "dev": true
     },
     "@material-ui/core": {
-      "version": "4.5.0",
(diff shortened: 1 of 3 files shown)


benmacleod (approved):
Storybook seems to work as before 👍🏻

PR #146, opened 2021-02-22
Switch build to use Node LTS 14

nickbrowne:
This should let us set up export mappings to optimise the imports for apps which use this library (not done yet).

Have tested build on Node 14 against TC and it appears to work fine.


ags (approved):
https://conversation.slack.com/archives/G02EEH7KF/p1613966458032100

PR #147, opened 2021-03-03
Create the Tab and export MUI's Tabs components

tiagoamaro:
**Why?**

We want to have access to MUI's tabbed navigation components

**Screencast**

tabbed-navigation-component


benmacleod (approved):
Provides a thin wrapper over MUI tabs, and minimal stories to illustrate their use 👍 

I approve. Does @ZoeJazz also need to approve?

tiagoamaro:
Nope @benmacleod , just adding the tab interface so we can use it through the UI lib instead of referencing `@material-ui/core`!

Any customization will happen at the component level, so it's really a thin wrapper!

Thank you!!

PR #169, opened 2021-07-13
Use MUI StylesProvider instead of react-jss JssProvider

nickbrowne:
They both perform the same function, but using MUI's built in StylesProvider should result in a smaller bundle for TC once we upgrade to npm 7.

Depends on https://github.com/conversation/tc/pull/11337

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4163,11 +4163,6 @@
         }
       }
     },
-    "brcast": {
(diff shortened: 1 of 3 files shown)


nickbrowne:
Confirmed that npm 7 is now working fine on TC, so this is unblocked.

benmacleod (approved):

PR #170, opened 2021-07-14
Relax lodash dependency version

tiagoamaro:
**Why?**

We don't want to force projects using this UI lib to have `lodash` version 4.17.21. Let's increase the version range to the patch version, using the caret syntax.


benmacleod (approved):
👍🏻

PR #174, opened 2021-09-09
Delete the Autocomplete component from the design system

tiagoamaro:
**Why?**

The `Autocomplete` component was never used.  It's clear that it fell into oblivion, as its `downshift` dependency was still fixed on version `3.2.7`, released three years ago.

Let's delete this component, simplifying the design system and allowing apps to have non-conflicting `downshift` versions.

**Manual testing steps**

- Checkout this branch
- Install modules (e.g.: `make node_modules`)
- Build storybook (e.g.: `make storybook`)

When you open storybook's server, you'll see that the `Autocomplete` component is gone.


RohanM:
Screenshot from 2021-09-10 13-46-25
 :tada:

RohanM (approved):
Looks good! If we decide we want an autocomplete in TC-UI in the future we can either bring this one back or decide on a fresh approach.

PR #175, opened 2021-09-10
Override MUI box shadow to lighten it

RohanM:
Align with brand guidelines for shadows - we'd like subtle shadows only.

### Before
Screenshot from 2021-09-10 11-33-30

### After
Screenshot from 2021-09-10 11-31-30

### DIFF
--- a/src/Button/Button.jsx
+++ b/src/Button/Button.jsx
@@ -9,6 +9,9 @@ const styles = theme => ({
     boxShadow: 'none',
     '&:active': {
       boxShadow: 'none'
+    },
+    '&:hover': {
+      boxShadow: '0px 2px 4px -1px rgb(0 0 0 / 10%), 0px 4px 5px 0px rgb(0 0 0 / 5%), 0px 1px 10px 0px rgb(0 0 0 / 5%)'
     }
   },
   prominent: {


src/Button/Button.jsx
```diff
@@ -9,6 +9,9 @@ const styles = theme => ({
     boxShadow: 'none',
     '&:active': {
       boxShadow: 'none'
+    },
+    '&:hover': {
+      boxShadow: '0px 2px 4px -1px rgb(0 0 0 / 10%), 0px 4px 5px 0px rgb(0 0 0 / 5%), 0px 1px 10px 0px rgb(0 0 0 / 5%)'
```
pawelgalazka, line 14: I wonder if we should switch to `hsl` color format, it is much easier to reason about color selection. Thoughts?

pawelgalazka: Some arguments in favour of HSL: https://sujansundareswaran.com/blog/why-hsl-is-better-than-hex-and-rgb

ZoeJazz: Happy either way. Wonder if the switch could make more sense when we update the buttons and move away from MUI?

ags: Let's go with rgb for now and look at adopting hsl more broadly in another change

ags (approved):

PR #185, opened 2021-11-25
Make sure that the package is built before it is published

nullobject:
There were a couple of issues with the new GitHub package workflow:

* We were publishing an empty package, this is because we forgot to build it first.
* We weren't uploading the storybook docs to S3

Both of these things have been addressed, and test packages have been built and deployed.

This also required creating a couple more organisation-level secrets for our AWS credentials (they are also in 1Password).

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -7,11 +7,28 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+
       - uses: actions/setup-node@v2
         with:
           node-version: '16.x'
           registry-url: 'https://registry.npmjs.org'
+
       - run: npm install
+
+      - run: npm run build
+
+      - run: npx build-storybook -c .storybook -o doc
+
+      - uses: jakejarvis/s3-sync-action@master
+        with:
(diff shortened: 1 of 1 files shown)


RohanM (approved):

.github/workflows/package.yml
```diff
@@ -7,11 +7,28 @@ jobs:
...
+      - uses: jakejarvis/s3-sync-action@master
+        with:
+          args: --acl public-read --delete --cache-control 'max-age=300'
+        env:
+          AWS_S3_BUCKET: 'styleguide.theconversation.com'
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          AWS_REGION: 'us-east-1'
+          SOURCE_DIR: 'doc'
+
```
tiagoamaro, line 31: `jakejarvis/s3-sync-action` GitHub action seems quite stable (with latest commit being on March 5, 2020, but would be interesting for us to point to a specific version instead of `master`?

PR #197, opened 2022-02-24
Remove storybook publication

ags:
We've deprecated this UI library and no longer want to make the storybook accessible, so this removes the actions to build and publish it.


benmacleod (approved):
Removes the GitHub action for building the now-redundant styleguide and syncing it to S3 👍🏻

PR #202, opened 2022-04-07
Pass id to Select in Dropdown

benmacleod:
With the upgrade to version 2.41.0 of this library, testing library in donations has effectively [started complaining](https://buildkite.com/conversation/tc-donations/builds/14481#7fcee22d-9468-4f7d-a293-f8e987799f72) that TC-UI `Select` elements do not have  ids (it can't find the associated form control for the label the spec is searching for, because the associated form control doesn't have the id we're passing in).

Given a `Dropdown` component with an id of `"xyz"`, the DOM should include a `select` with an id of `"xyz"` as well as a label with a `for` attribute of `"xyz"`.

### Testing ###

- Checkout this branch
- Run `make storybook`
- Check the Dropdowns > States page (e.g. http://localhost:9001/?path=/story/dropdowns--states) and note that the `select` elements in the DOM have the ids specified in the associated [story](https://github.com/conversation/ui/blob/39da790fd42f0c7b5f15e9573227f74b53dbe16e/src/Dropdown/stories/states.jsx#L59):

Screen Shot 2022-04-07 at 1 37 17 pm

### DIFF
--- a/src/Dropdown/Dropdown.test.jsx
+++ b/src/Dropdown/Dropdown.test.jsx
@@ -23,10 +23,11 @@ describe('<Dropdown />', () => {
 
   it('passes props to the Select component', () => {
     const wrapper = shallow(
-      <Dropdown value='foo'>
+      <Dropdown id='abc-1' value='foo'>
         <option value='1'>one</option>
       </Dropdown>
     ).dive().find(Select)
+    expect(wrapper.props().id).toBe('abc-1')
     expect(wrapper.props().value).toBe('foo')
   })
 
--- a/src/Dropdown/Dropdown.jsx
+++ b/src/Dropdown/Dropdown.jsx
@@ -79,6 +79,7 @@ export const Dropdown = ({
         </FormLabel>
       )}
       <Select
+        id={id}
         {...other}
         input={InputElement}
         MenuProps={{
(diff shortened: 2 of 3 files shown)


src/Dropdown/Dropdown.test.jsx
```diff
@@ -23,10 +23,11 @@ describe('<Dropdown />', () => {
 
   it('passes props to the Select component', () => {
     const wrapper = shallow(
-      <Dropdown value='foo'>
+      <Dropdown id='abc-1' value='foo'>
         <option value='1'>one</option>
       </Dropdown>
     ).dive().find(Select)
+    expect(wrapper.props().id).toBe('abc-1')
     expect(wrapper.props().value).toBe('foo')
```
tiagoamaro, line 31: Nit: we could use the `.toMatchObject` Jest matcher, but really, a nit.

```suggestion
    expect(wrapper.props()).toMatchObject({
      id: 'abc-1',
      value: 'foo'
    })
```

tiagoamaro (approved):
LGTM!

The HTML `label[for]` attribute value now has a `select[id]` target to match with!

tiagoamaro:
@benmacleod , just marked this fix with the patch version `v2.41.1`. It should be pushed on `npm` in a few minutes.

PR #204, opened 2022-04-27
Use Artifact Registry base image

ags:
See:
* https://github.com/conversation/docker-images/pull/26
* https://github.com/conversation/docker-images/pull/27

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM theconversation/node-ubuntu:16.13.1
+FROM us-docker.pkg.dev/tc-platform-artifacts/theconversation/node:16.14.2
 
 # Add files required for storybook
 COPY package*.json /app/


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM theconversation/node-ubuntu:16.13.1
+FROM us-docker.pkg.dev/tc-platform-artifacts/theconversation/node:16.14.2
```
jameshill, line 1: i'm getting confused...

isn't it node-ubuntu which is 16.14.2?

jameshill: or this dependent on https://github.com/conversation/docker-images/pull/29/files

jameshill: have now merged https://github.com/conversation/docker-images/pull/29 so 👍

PR #206, opened 2022-04-29
Use node 16.15.0

ags:
This upgrade is mostly for consistency with other apps, which upgraded as part of addressing an issue installing node from nodesource while building development docker images.

See:
* https://github.com/conversation/docker-images/pull/31
* https://github.com/nodejs/node/blob/v16.15.0/doc/changelogs/CHANGELOG_V16.md


jameshill (approved):
👍

PR #231, opened 2024-03-22
Fix issues for maintainbility

RohanM:
When responding to dependabot PRs, I found a few issues with this repository:

1. There was no `.node-version` set, and using the same version as TC prevented storybook from running
2. The package lock was out of date and needed a one-time update
3. The example react app was broken (on either version of node), appeared unused, yet was producing dependabot updates.

Let's set the node version to the same version set in the Dockerfile, update the package lock and remove the example app.


RohanM:
[Set node version and update package lock](https://trello.com/c/1l6DczW8/16613-set-node-version-and-update-package-lock)

README.md
```diff
@@ -102,16 +102,6 @@ const App = () => (
...
-## Examples
-
-To run the example react app:
-
-```sh
-> cd examples/react-app
-> npm i
-> npm start
-```
-
```
RohanM, line 114: Happy to hear if this example app provides value that I'm not aware of, in case it might be worth keeping/maintaining.

ags (approved):

README.md
tiagoamaro: I does not add any value that I'm aware of @RohanM. I cannot see it being more valuable than Storybook, so I think we could :fire: it.

